PR #12, opened 2019-08-19
Run eslint in travis; fix linting errors

talwrii:


willrogers:
npm managed to show this rather detailed message (when I had the wrong version). 

I think the upshot is that we're not supposed to pin the version of eslint and create-react-app will take care of the version.

```
will@wrmpb ~/c/cs-web-proto> npm test

> cs-web-proto@0.1.0 test /Users/will/code/cs-web-proto
> react-scripts test --coverage


There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "eslint": "^5.16.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of eslint was detected higher up in the tree:

  /Users/will/code/cs-web-proto/node_modules/eslint (version: 6.2.0) 

Manually installing incompatible versions is known to cause hard-to-debug issues.

If you would prefer to ignore this check, add SKIP_PREFLIGHT_CHECK=true to an .env file in your project.
That will permanently disable this message but you might encounter other issues.
```

talwrii:
Cool cool. I'm not sure it's worth caring that much about this - so I've reverted the pin.

willrogers:
Thank you

PR #14, opened 2019-08-20
Prettier wanted some changes on my machine.

willrogers:
But Travis said the previous version was ok. What's going on?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ node_js:
   - 11
 script:
   - npm test
-  - npx prettier-check **/*.ts[x]
+  - npm run prettier-check
   - npm run coveralls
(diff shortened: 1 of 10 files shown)


willrogers:
@talwrii Travis didn't seem to be running things correctly. When I got eslint running it didn't like a couple of things so I had to do some rearranging to avoid the non-null-assertions. See what you think.

willrogers:
Yeah that's just a relic of my hacking. The URLs were actually hard-coded in `coniql.ts`; now at least they're hard-coded in `App.tsx`.

willrogers:
Does it work now? I'd have to actually run up the Coniql server. I would like more unit tests instead.

PR #17, opened 2019-08-21
BaseWidget, progressbar and restructuring

TimGuiteDiamond:
BaseWidget added which provides useful functionality across all widgets.
Further refactoring required to maintain composable design philosophy - functionality should be separated out to container components which may or may not be used but can be tested individually.

ProgressBar widget added as demonstration and comparison to CS-Studio widget. Looks nice.

Components all in their own folders.

index.html updated to include new widget layout.


willrogers:
Some further changes to come as discussed.

Issue #18, opened 2019-08-21
Composable Components


TimGuiteDiamond:
Refactor use of container widgets (such as BaseWidget) to allow composable design style.

This will allow flexibility in use of containers as well as the capability to test components more easily.

willrogers:
This is now working as intended. There are related questions that can have their own tickets.

PR #20, opened 2019-08-21
Refactored Readback wiget

TimGuiteDiamond:
Removed BaseWidget wrapper, added precision, alarm input and styling.

Updated readback and progressbar to use new csState file. Also amended progress bar to use local pv.


TimGuiteDiamond:
Working towards #18

PR #23, opened 2019-08-22
Connections

willrogers:
This is still pretty untidy but I thought I would push it for discussion.

It adds a connection status to the store, separate to the latest value for a PV.

It also pulls out the action of connecting into a wrapper component. I think that wrapper should pass an `NType` plus a connection state into each component but at the moment it passes a string.

I updated the simulator a bit to add a PV that periodically disconnects. We should really extend it further so we can get out alarm status etc.

Comments welcome @TimGuiteDiamond @talwrii


willrogers:
It's difficult to balance heavy refactoring with writing tests... I'm not sure how much it's worth writing lots of tests before we are happy with APIs.

TimGuiteDiamond:
I think that this approach makes sense for the PV connection, as it is simply wrapping the components with some particular data.

It seems to me that this is different from the CopyWrapper in that it does not actually add any html code to the final DOM. The copy wrapper will add a div and a span and we want this to be reflected in the final DOM output. Therefore, perhaps it makes more sense to keep it as a pure component.

This would lead to a standalone readback looking like, for example:
```
const standaloneReadback = (props) => (
    <div>
        <CopyWrapper pv={props.pv}>
            <Readback pv={props.pv} />
        </CopyWrapper>
    </div>
)
```
and a connecting it:
```
export const connectedStandaloneReadback = connectionWrapper(standaloneReadback)
````

This seems relatively clean to me - pure components expect a normative type and render themselves appropriately, regardless of where they sit in the hierarchy etc.

willrogers:
This is interesting. I had come to the conclusion that we needed higher-order compnents (see [here](https://reactjs.org/docs/higher-order-components.html)) but I don't think that's actually the case. What Tim describes is just a container component. I think in our cases we can use container components so I will look at refactoring to use them.

willrogers:
After a bit more thinking I'm not sure it's possible to do anything in Tim's example that changes the props being passed down. As long as all the props are passed through to the child component (as should be the case in the `CopyWrapper`) it looks like it should work fine, although I've still got a TypeScript error I don't understand.

willrogers:
I'm going to merge this, we can fix any problems we find later!

Issue #10, opened 2019-08-19
Detect import cycles


willrogers:
Import cycles typically suggest some design flaw. `tslint` will detect them for you but is deprecated in favour of `eslint`, which doesn't seem to do so.

PR #25, opened 2019-08-27
Added eslint-plugin-import to prevent circular dependencies

TimGuiteDiamond:
Addressing #10

Plugin found here: https://github.com/benmosher/eslint-plugin-import

Advice found here: https://railsware.com/blog/how-to-analyze-circular-dependencies-in-es6/

Plugin use described in package.json, this could be moved to .eslintrc.yml or similar at a later point


willrogers:
I suggest that you create a circular dependency in one commit to make sure that Travis fails, then correct it in another. Then we can be confident that this check is working.

TimGuiteDiamond:
Already found circular dependencies in this project with this rule, suggesting it works. Project was running fine but bad practice to have these circular dependencies hanging around.

Clean up underway before merging.

PR #26, opened 2019-08-27
New wrapped components

TimGuiteDiamond:
"Standalone" readback and progress bar components presented which are wrapped in copy border and alarm borders.

Demonstrate composability of project so far.


willrogers:
There's plenty of naming questions here, such as the difference between a `ConnectedReadback` and a `ConnectedStandaloneReadback`, but the principle works. Let's merge and move on.

PR #27, opened 2019-08-28
Added redux devtools functionality

TimGuiteDiamond:


willrogers:
The title says 'Added redux devtools functionality' but this does a lot more than that. Can you describe it?

willrogers:
It looks like you only wanted the redux commit but you got everything else as well. You probably want to force-push the `redux` branch with only that one commit.

TimGuiteDiamond:
From the commit message:

Followed instructions at https://github.com/zalmoxisus/redux-devtools-extension to connect to redex devtools extension in browser development tools.

Allows view, pause, rewind of redux state for debugging.

Added window as any type following advice on the answer from Dimitar Nkoviski on this Stack Overflow question: https://stackoverflow.com/questions/12709074/how-do-you-explicitly-set-a-new-property-on-window-in-typescript

TimGuiteDiamond:
This basically allows you to view the state of the redux store at any point in time and move forwards and backwards to see what effect changes in state have on the DOM

Issue #36, opened 2019-09-06
Conventions


willrogers:
We are using Prettier, but we still have quite a few irregularities in the code, filenames being one.

I suggest we follow https://github.com/basarat/typescript-book/blob/master/docs/styleguide/styleguide.md, and make any further updates to [this wiki page](https://github.com/dls-controls/cs-web-proto/wiki/Style-guide).

https://www.npmjs.com/package/eslint-plugin-filenames may help.

TimGuiteDiamond:
Cannot work out how to properly implement this eslint extension, does not seem to want to behave. Will try again on Monday.

TimGuiteDiamond:
Figuring this out took me quite a while@:
https://github.com/eslint/eslint/issues/9330

So the glob expression will now be wrapped in single quotes

PR #39, opened 2019-09-16
Added filename linting to provide consistent camelCase file names

TimGuiteDiamond:
Added quotes to lint glob to avoid inconsistencies in which files are linted. This should help with other rules as well.

Added the eslint-plugin-unicorn package as their filename linter is better than eslint-plugin-filename (it detects *.test.tsx files properly). Also contains a multitude of other rules which could be useful for us.

Specifically disabled this eslint rule in App.tsx, App.test.tsx, react-app-env.d.ts which are files set up by create-react-app. Used the mechanism described here: https://eslint.org/docs/user-guide/configuring.html#configuring-rules


TimGuiteDiamond:
Works on #36, recommend merging this is and changing filenames as the last in a series of merges, otherwise merging the other branches will be a nightmare

PR #43, opened 2019-10-02
Change input behaviour on enter

skyfrench:
Tiny tweak which just changes how the text input boxes behave after pressing return (editing is set to false so the new input value is rendered, not a blank value).


willrogers:
Thank you.

PR #45, opened 2019-10-04
Theme demo

skyfrench:
A possibly not very useful demo of toggling a site-wide 'theme'.


willrogers:
Is this one still valid?

The prettier (code auto-format) problems should be solved by setting up prettier to run on save in VS Code. There is a brief mention on this page: https://github.com/dls-controls/cs-web-proto/wiki/Development-Environment

skyfrench:
@willrogers - I've merged this into the styling branch and updated it to use CSS variables to be compatible with the styling approach implemented there.

PR #50, opened 2019-10-09
Macros

willrogers:
A slightly different design with less in the store and substitutions happening at render time.


TimGuiteDiamond:
This is looking good with me, merged with latest changes to master branch. Ready to go @willrogers

PR #51, opened 2019-10-09
Positioning

TimGuiteDiamond:
Changes to provide flexible layouts and embedding of screens from multiple files.

Also amended issue with borders not adjusting to size of widget when lacking size specifications of parent component.


src/components/FlexContainer/flexContainer.module.css
```diff
@@ -0,0 +1,10 @@
+.FlexContainer {
+  display: -webkit-box;
```
willrogers: I don't think we need these.

src/components/FromJson/fromJson.tsx
```diff
@@ -1,12 +1,17 @@
-import React, { useState } from "react";
+import { useState } from "react";
```
willrogers: We can rename this file to `.ts`.

src/components/Positioning/positioning.tsx
```diff
@@ -11,13 +11,33 @@ export interface PositionDescription {
...
+  // Object for styling the container which contains the children
+  containerStyling?: object;
+  // Array of any children nodes - children are all at same level
+  // with respect to positioning
+  children?: (PositionDescription | FlexiblePositionDescription)[] | null;
 }
 
 export function objectToPosition(
-  inputObjects: PositionDescription | null,
+  inputObjects: PositionDescription | FlexiblePositionDescription | null,
```
willrogers: `AbsolutePositionDescription`?

src/components/Positioning/positioning.tsx
```diff
@@ -49,18 +71,34 @@ export function objectToPosition(
...
-          left: x,
-          top: y,
-          width: width,
-          height: height
-        }}
-      >
-        <Component {...otherProps}>{PositionedChildren}</Component>
-      </div>
-    );
+    if (flexible === false) {
```
willrogers: Just create the correct style object then have only one return?

PR #56, opened 2019-10-11
Enum

willrogers:
A VEnum and sim PV.


TimGuiteDiamond:
Made a change to the simulator, let me know what you think.

Also, when you try and change the value of an actual EPICS enum record, do you send the index, the value you want, or a whole new PV? Will research this further if not well known.

TimGuiteDiamond:
@willrogers added functionality to VEnum, does this look sensible? Alternative was to chain commands as: value.getDisplay().getChoices()

PR #59, opened 2019-10-15
Vtypes

willrogers:
Fairly innocuous but will cause a few conflicts @skyfrench @TimGuiteDiamond


TimGuiteDiamond:
Why is string separated from VTypes or will this be done the the other types as well?

willrogers:
Probably, yes.

TimGuiteDiamond:
What is the benefit of that over having all the VTypes in one file?

willrogers:
It's going to be a very big file.

PR #60, opened 2019-10-15
Coverage

willrogers:
@skyfrench @TimGuiteDiamond minor changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,10 @@ licensed with the Apache License: see `LICENSE`.
 
 ## Setup
 
+You will need NodeJS. See
+[this wiki page](https://github.com/dls-controls/cs-web-proto/wiki/Development-Environment)
+for installation instructions.
+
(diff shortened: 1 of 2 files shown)


talwrii:
Cool cool. Should make output a bit more readable. 

I'm going to mutter about what I've found about running tests here... because this seems like a relevant place.

I'm running tests like so:

`CI=true npx react-scripts test  /home/tom/work/cs-web-proto/src/connection/sim.test.ts -t 'local values`

The CI=true test stops the watcher / rerunner - which can take up screen space;  `-t` allows one to specify an individual test by a substring (pattern matcher?) for the first argument to it.

There is supposedly a jest plugin for vscode (which I don't use) and also a jest plugin for emacs (which I do use). The jest plugin for emacs appears to not work because react-scripts is doing some wrapping around jest. Hence the `npx react-scripts` command.

Issue #35, opened 2019-09-06
Simulation Upgrade


TimGuiteDiamond:
The simulation is reaching its limits with PVs that can be initialised and reinitialised, have limits and timestamps etc. There is also a need to add enums and arrays to the simulation.

In order to tie in with #34 it seems that an upgrade may be required. It would be nice to initialise a simulation environment of some kind from JSON which communicates through the middleware in the same was the Coniql client or other middleware.

Therefore, the simulation should:
- be initialised separately from the components rendered on screen
- maintain its own values separately from the store
- provide a mechanism for updating values
- provide a mechanism for pushing values on changes or new subscriptions
- provide a mechanism to stop pushing values when the subscription has finished

In effect, we should aim to more fully simulate the environment we wish to use the application in.

talwrii:
I want to split out the separate changes into sections.

# Remove namespaces

meta:// lives in a separate namespace. We could replace this with sim://limits

# Make some simulated values writeable

`meta://` is already writeable. If we are to move `meta://` in to the simulated namespace then it seems to make sense to make most simulated pv's writeable (with the exception of those that generate random values).

This could be achieved by having SimPv implement an updateValue method that can error out for e.g. sine simulated pv's.

willrogers:
I just had a talk to Tim. We could potentially have multiple `sim://limits` PVs (e.g. `sim://limits100`) if we felt it was necessary. But sims aren't fundamental, they are for demonstration purposes, and something like `meta://` isn't that flexible for other purposes.

willrogers:
I don't think there's any need to initialise PVs other than at the time they are requested. But it might be nice to close them when the last subscriber is removed, as for real PVs.

talwrii:
> We could potentially have multiple sim://limits

I might vote for sims://limits#name so as to ape html; or perhaps sims://limits.name

willrogers:
That's an interesting idea.

TimGuiteDiamond:
The "initialise separately" part was looking towards initialising a simulated database from a JSON file or similar and then having the client connect to that. This would be useful because for defining multiple "meta" type PVs each with different limits.

However, that in itself might not be that useful for demonstration and perhaps there are other ways to do this on the fly based on the PV name, the # idea seems interesting to me, how much information do you think could be squeezed into something like that? PV type, limits, readonly?

willrogers:
[...]

PR #63, opened 2019-10-17
Talwrii  issue 35  simulator refactor

talwrii:
This implements various parts of #35.  Specifically it

* Replaces meta:// with sim://limit
* Removes quite a lot of duplicated code
* Adds tests

What exactly we want to do in the #35 is still kind of amorphous, so it doesn't quite close this issue. This seems a large enough change that it wants to get merged however.

### DIFF
--- a/src/connection/coniql.ts
+++ b/src/connection/coniql.ts
@@ -19,12 +19,12 @@ import {
   nullConnCallback,
   nullValueCallback
 } from "./plugin";
-import { PartialVType } from "../redux/csState";
 import { VType } from "../vtypes/vtypes";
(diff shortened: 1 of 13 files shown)


willrogers:
We should remove the `enum://` prefix in favour of `sim://enum`.

willrogers:
Also there are some `meta://` PVs hanging around that we should clean up. I'd like to get this merged soon though.

willrogers:
Thinking about it, there are two enum PVs here. The readonly one can be `sim://enum`. The writeable one should be `loc://xxx<VEnum>(0, "zero", "one", "two")`.

Issue #55, opened 2019-10-10
Rules


willrogers:
Rules allow changing any property based on the value of different PVs. They are encoded in the `.bob` file format.

This should be possible with a wrapper component. The type of the property is important.

TimGuiteDiamond:
Will this be able to change the name of the PV which the element is interested in? If so might it need to be a HOC wrapped around macro and connection wrappers?

willrogers:
So, I have thought about wrapper components and props and higher-order components.

The difficulty with wrapper components is that you can't edit the props that children get when you render `{props.children}`. This works ok when you are just _using_ props (like in the PV copy, which uses the PV name, and the tooltip, which uses the name and value). It doesn't work for the connection wrapper, which requires a PV name but needs to add the value as a prop, or the macro wrapper which needs to change the values of props.

[This blog post](https://frontarm.com/james-k-nelson/passing-data-props-children/) among others describes this problem.

The way I have managed this is to use the so-called 'higher-order components'. These take the component as an argument and then render that component with any additions or changes to the props. Since the natural way to define rules modifies the props for children we should probably use a similar HOC. See `src/Components/ConnectionWrapper/connectionWrapper.tsx`.

@skyfrench @TimGuiteDiamond

willrogers:
Here's a snippet from a Phoebus XML file. The rule is the one that changes x position based on the value of the PV.

```
    <rules>                                                                     
      <rule name="xpos" prop_id="x" out_exp="true">                             
        <exp bool_exp="true">                                                   
          <expression>pv0*100 + 100</expression>                                
        </exp>                                                                  
        <pv_name>sim://sine</pv_name>                                           
      </rule>                                                                   
    </rules> 
```

TimGuiteDiamond:
Added in #76

TimGuiteDiamond:
Only allows individual rules now, not multiple rules

willrogers:
We now have a more sophisticated rules mechansim.

PR #64, opened 2019-10-18
WIP: rules

skyfrench:
A little progress on #55 for discussion ... particularly how it might change given @TimGuiteDiamond's work on a widget interface.

### DIFF
--- a/src/components/RuleWrapper/ruleWrapper.tsx
+++ b/src/components/RuleWrapper/ruleWrapper.tsx
@@ -41,7 +41,7 @@ export const RuleWrapper = <P extends object>(
       valid = false;
     else {
       let pvs = [];
-      for (let [name, pv] of Object.entries(props.substitutionMap)) {
+      for (let [pv] of Object.entries(props.substitutionMap)) {
         pvs.push(pv);
       }
       // eslint-disable-next-line react-hooks/rules-of-hooks
--- a/src/components/RuleWrapper/ruleWrapper.tsx
+++ b/src/components/RuleWrapper/ruleWrapper.tsx
@@ -59,11 +59,12 @@ export const RuleWrapper = <P extends object>(
         } else valid = false;
       }
     }
-    if (valid && condition !== undefined) {
+    if (valid && condition !== undefined && props.prop !== undefined) {
       let state = evaluate(condition);
       let styleValue = state ? props.trueState : props.falseState;
-      // use props.prop not hard coded colour
-      return <Component {...(props as P)} colour={styleValue}></Component>;
+      const newProps = { ...props };
+      newProps[props.prop] = styleValue;
+      return <Component {...(newProps as P)}></Component>;
     }
     return <Component {...(props as P)}></Component>;
   };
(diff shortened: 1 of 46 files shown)


TimGuiteDiamond:
This is pretty cool actually! 

I think that it could be quite powerful if wrapped around Widget but yes it would need a bit more intelligence in order to correctly separate out containerStyling vs widgetStyling for example.

willrogers:
Looks nice. Two points:

* `useSubscription` makes it difficult or impossible for you to conform with the hooks rules. Perhaps I could make it take a list of PVs, then you could figure out the ones you want and call from the top-level.
* we should put `useId()` into the top-level `Widget` component when that exists properly, then pass it down as a prop.

willrogers:
Note the ID is per component not per subscription so you could get your ID at the start of `RuleWrapper` and remove one of the warnings. But what I said above applies once we have a Widget component.

I just pushed a `subs` branch that might help with the other warning, it changes `useSubscription` to take an array of PVs.

src/components/RuleWrapper/ruleWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+        });
+        if (latestValue !== undefined && connected) {
+          condition = condition.replace(name, latestValue.getValue());
+        } else valid = false;
+      }
+    }
+    if (valid && condition !== undefined) {
+      let state = evaluate(condition);
+      let styleValue = state ? props.trueState : props.falseState;
+      // use props.prop not hard coded colour
```
willrogers: Something like

```
const newProps = { ...props };
newProps[props.prop] = styleValue;
return <Component {...(newProps as P)}></Component>;
```

?

skyfrench: That works :)

willrogers:
It now looks to me like we can't call a hook inside a HoC and still obey the 'rules of hooks'.

Are we going to need a refactor? Can we make our HOCs into hooks? I find it hard to think about these things clearly.

willrogers:
You're right I am apparently making the same violation in `connectionWrapper` but I don't get an error. That is weird.

willrogers:
Er... if you change the name from `RuleWrapper` to `ruleWrapper` then two of your errors go away. I'm not sure what to think of this but it's something to do with the wrappers not being components, but returning components.

It still leaves a problem with calling `useSelector` in a loop.

src/components/RuleWrapper/ruleWrapper.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return (props: RuleProps): JSX.Element => {
+    // eslint-disable-next-line react-hooks/rules-of-hooks
+    const [id] = useId();
+    let condition = props.condition;
+    let valid = true;
+    if (condition === undefined || props.substitutionMap === undefined)
+      valid = false;
+    else {
+      let pvs = [];
+      for (let [pv] of Object.entries(props.substitutionMap)) {
```
willrogers: This is picking the key not the value. You could use `Object.values()`

willrogers:
n.b. looks like there are some problems on the `subs` branch that I need to resolve. I'm getting repeated subscriptions and unsubscriptions.

willrogers:
I have pushed another commit to `subs` that resolves the problem I was referring to.

willrogers:
This is now adapted and merged in the `hooks` branch so I'll close this.

PR #67, opened 2019-10-21
Widgetinterface

TimGuiteDiamond:
Added a new Widget component which describes a general Widget and has a base interface which all widgets are expected to follow.

"Widgified" some existing components and added some new components as widgets. In the future, all components will be replaced with widgets, with the exported widget having a simple name - Readback, Input etc - and the lower level components having more complex names as they are for developers.

Have also provided new json files for all screens apart from the macros screen.

All functionality is preserved. Absolute and relative layout are supported with the use of flexboxes.


TimGuiteDiamond:
@willrogers pretty close on lines added vs removed! Might add some tests for widget though and will merge with master

TimGuiteDiamond:
I think test coverage has decreased because the components are all genereated from JSON rather than instantiated directly.

Would it be worth having one page where we directly render every widget we have created thus far purely for test purposes?

Or otherwise is it possible to mock out the calls to the store etc? For the purposes of rendering the widget? Otherwise we may just have to accept this test coverage

TimGuiteDiamond:
Or maybe it's ok? :man_shrugging:

willrogers:
This is nice. We have moved a lot of JavaScript to JSON and the Widget component isn't too complicated.

Some future improvements:

* make the JSON closer to the existing XML formats
* move wrappers and components into different directories
* copywrapper does middle-click copy and tooltip
* middle-click copy should be default for a pv widget
* border and alarm border should probably be separate
* probably some renaming

I'm going to merge this now.

PR #70, opened 2019-10-23
phoebus local value initialization

talwrii:


src/components/ConnectionWrapper/connectionWrapper.tsx
```diff
@@ -7,22 +7,32 @@ import { VType } from "../../vtypes/vtypes";
...
+export interface PvErrorProps extends PvProps {
+  error: boolean;
+}
+
+
 function pvStateSelector(
   pvName: string,
   state: CsState
-): [boolean, boolean, VType?] {
+): [string, string | undefined, boolean, boolean, VType?] {
```
willrogers: `string | undefined` is `string?`

willrogers:
I don't think you committed the `processVariable.tsx` file.

willrogers:
I just tested on Phoebus and it seems that the first local PV to be initialised wins. That appears to be the first one in the XML file, but I'm not certain that that's deterministic.

talwrii:
Okay this should be done.

src/connection/sim.ts
```diff
@@ -356,81 +352,153 @@ export class SimulatorPlugin implements Connection {
...
-    } else if (parts[0] === "sim://enum") {
+      initial = undefined;
+    } else if (nameInfo.protocol === "sim://enum") {
       cls = SimEnumPv;
-    } else if (pvName.startsWith("enum://")) {
-      cls = EnumPv;
-    } else if (parts[0] === "sim://random") {
+      initial = undefined;
+    } else if (nameInfo.protocol === "enum://") {
+      return { simulator: undefined, initialValue: undefined };
```
willrogers: We don't need this special case for `enum://`, just use the default.

PR #73, opened 2019-10-25
Actions

willrogers:
@TimGuiteDiamond @skyfrench action button and actions


skyfrench:
Can we merge this in (once conflicts are resolved)? It'd be useful to build on it in #101.

willrogers:
This mostly looks fine. What's the difference between an action button and a basic button?

Also we'll need to prop-types-ify the widgets, then we can merge.

skyfrench:
An action button you click and it will always perform an action (or actions).

A basic button is simply image or text which can be wrapped to have e.g. a context menu, or illustrate a status PV.

willrogers:
Could we merge them so that if an action button doesn't have an action it behaves like a basic button?

willrogers:
This for a future pull request I think:

I think we want to have our context menu wherever you click.

Could we put this in at some high level but when the click is intercepted determine which component (or components) is (or are) the target?

PR #76, opened 2019-10-25
Hooks

willrogers:
@skyfrench @TimGuiteDiamond from HOCs to hooks.

I don't think this really helps one way or another with Sky's problem, so the question is whether it's an improvement it its own right. I think it probably is.


src/components/Widget/widget.tsx
```diff
@@ -146,10 +148,6 @@ export interface WidgetInterface extends BaseWidgetInterface {
   children?: React.ReactNode;
 }
 
-export const Widget: React.FC<
-  WidgetComponentInterface & WidgetInterface
-> = macroWrapper(WidgetComponent);
-
```
TimGuiteDiamond: Why has this been deleted?

TimGuiteDiamond: Does renaming WidgetComponent to Widget actually work? Surely it causes problems for non PV widgets which will require the baseWidget at runtime? Perhaps this is not causing errors from JSON...

willrogers: I think we can do some further cleaning up of these interfaces. We can talk about it next week.

TimGuiteDiamond:
The widget test needs clearing up because it now looks quite different.
Could do with testing recursiveWrapping separately as this is what is really being tested in those tests.

Will direct testing of Widget and PVWidget no longer be possible due to the use of hooks? Perhaps they can be mocked out?

willrogers:
This now includes a `useRules` hook adapted from Sky's `rulesWrapper` and conforms with the rules of hooks. I've put in enough testing for Coveralls to be happy, but there's certainly more that could be done.

I think this is an improvement so I'll merge when you've had a chance to comment.

TimGuiteDiamond:
Can we apply multiple rules using this sturcture?

willrogers:
Not yet.

PR #77, opened 2019-10-26
Logging -- Standardize log levels and document usage

talwrii:
I have found the debug log level particularly useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,3 +46,18 @@ For more details on the development environment see
 - [Jest](https://github.com/facebook/jest) is a widely-used testing framework
   developed by Facebook
 - [Prettier](https://github.com/prettier/prettier) is used for code formatting
+
+
+## Debugging
+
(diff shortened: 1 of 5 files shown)


willrogers:
We don't want `trace` by default do we?

willrogers:
Also I expect there's room for using the `warn` level.

willrogers:
Too late I added a commit and merged.

talwrii:
> We don't want trace by default do we?

Fixed

> Also I expect there's room for using the warn level.

Sure.  Can you think of anything to warn about : )?

I graded the warning into levels... and only really needed four levels. I could imagine using warn for "important info". We might have components be quite chatty about internal state changes on info mode, but only show the importation messages in info mode. But then... if we are showing detailed information from different components we might want to be able to tweak logging per component. 

----

Mutter in general about use cases for logging:

i. Isolating observered behaviour to a particular component: info
ii. Checking what messages to the redux stream look like: debug I was using this to check that process variable names were correct. I imagine it might be useful to filter this
iii. Checking what messages from the network look like
iv. Adding manual console.log lines to a component and checking that they do the correct thing after a particular message.

Issue #69, opened 2019-10-21
Runtime type checking


TimGuiteDiamond:
It would be good to be able to validate data coming from JSON to guarantee that the necessary props are being passed to widgets.

This requires runtime validation against types/interfaces which, unfortunately, appears to be *very* difficult in typescript. It is a "non-goal" of the language (https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals non-goal 5).

The solutions vary from including a "type" field in all interfaces which may only have one value (https://www.matthewgerstman.com/tech/ts-tricks-type-guards/) this seems insufficient for us.

Then there are many middlewares which offer to do both static and runtime type checking, the library of the day seems to be io-ts (https://github.com/gcanti/io-ts). This would not be a minor change and would represent a relatively significant reliance on a library which may or may not continue to be supported moving forwards.

I haven't found any other great ideas other than very manual processes so any help would be much appreciated.

willrogers:
Yes, I was thinking that this would happen. Of course, if we try to use the TypeScript types then we can't support plain JavaScript widgets either. So perhaps that's not the best way.

We might be able to go the other way round. You can do type checking on components in JavaScript using [proptypes](https://www.npmjs.com/package/prop-types). You then seem to be able to infer the TypeScript type from the prop types using [inferProps](https://dev.to/busypeoples/notes-on-typescript-inferring-react-proptypes-1g88). And I think you can get at the prop types at runtime for e.g. editor purposes (e.g. [parse-prop-types](https://github.com/diegohaz/parse-prop-types)).

It's not fully clear that all the pieces we need are there, but they certainly might be.

TimGuiteDiamond:
Just had a quick look now but seems like a react-specific flavour of the above libraries? Perhaps that will be enough.

Could support JS by requesting an interface type object alongside the widget and if there isn't one then allowing all props through without an error?

talwrii:
Some thoughts:

* When do we want to error to occur? One thing we might want to do here is ensure that the error happens immediately on passing the errors. What does this mean for doing the checking on components? This might be okay if we create the components immediately on loading the JSON.
* Clear error messages. If we have lots of similar objects we want to be able to point the user / developer at the correct nested JSON object and potentially the correct line. One way of achieving this is to wrap our rendering components with wrappers that will print the path to an element on error.

TimGuiteDiamond:
[...]

PR #84, opened 2019-10-28
Prop types

TimGuiteDiamond:
Only a basic demo so far but demonstrates using prop-types to perform runtime type checking.

Might suggest taking https://github.com/ratehub/check-prop-types and converting to TS and including directly in our project, perhaps this would be more trouble than it is worth? Provides an actual functional interface to prop type checking.


TimGuiteDiamond:
Merged with latest changes from #76 

Showing build errors due to the checking of PropTypes in a foreign file when TS can't be assured that this attribute will exist. Fair enough but if we were to implement this the error could be disabled for this line.

Also because I am throwing an object as an error, perhaps there is a more graceful way to extract this information, particular to the widget and file but could be hidden within recursive calls?

TimGuiteDiamond:
Actually caught some errors in #76 FrontPage.json and would have helped @skyfrench (and other developers) when developing #92

TimGuiteDiamond:
See #69 for more information

willrogers:
The commit I pushed removes `null` from the inferred proptypes, which simplifies the code again.

I also added some ideas regarding reducing the verbosity of the proptypes code (i.e. type aliases), but they're no big deal either way.

TimGuiteDiamond:
A lot of the extra lines come from the verbosity forced by PropTypes, however as mentioned this is catching errors even with us so is probably worth keeping in for the time being.

PR #85, opened 2019-10-29
Context menu

skyfrench:
@TimGuiteDiamond @willrogers context menu to work with actions branch


willrogers:
That doesn't look too complicated. Two ideas from a quick test:

* persuade it to a similar width to the firefox context menu. It's a bit squished at the moment
* we need it to cancel if you click outside the menu, I'm not sure how easy that is

skyfrench:
> That doesn't look too complicated. Two ideas from a quick test:
> 
> * persuade it to a similar width to the firefox context menu. It's a bit squished at the moment
> * we need it to cancel if you click outside the menu, I'm not sure how easy that is

Yes - at the moment I've cheated and it closes if you right click again. I'll keep trying ...

skyfrench:
Added to #73

PR #87, opened 2019-10-29
Fixes

### LINKED ISSUE #86: Crash loading ConiqlPlugin
When you start using the ConiqlPlugin with coniql running, frequently you get a crash.

Somehow a VType with value null has arrived. We could fix this in `vtypeToString()`, but it would be nice to understand what's happening.

cc @skyfrench @TimGuiteDiamond 

type-error

willrogers:
Closes #86.

There's one more change I want to make on this branch.

@TimGuiteDiamond @skyfrench

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,7 +6,6 @@
   },
   "extends": [
     "react-app",
-    "eslint:recommended",
     "plugin:react/recommended",
     "plugin:@typescript-eslint/recommended",
     "prettier/@typescript-eslint",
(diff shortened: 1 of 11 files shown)


willrogers:
This branch adds a new check in Travis. You can run it locally with 

```
npx tsc
```

It runs the TypeScript compiler, but unlike `npm start` it runs it on the tests as well. This catches some problems that we had been missing. (They are shown for me in VS Code if I'm looking at the file.)

PR #90, opened 2019-10-30
Simulator -- Persist simulated values after unsubscribe

talwrii:


src/connection/sim.ts
```diff
@@ -471,19 +484,18 @@ export class SimulatorPlugin implements Connection {
         this.simPvs.put(simulatorInfo.simulator);
       }
 
-      if (publish && simulatorInfo.simulator !== undefined) {
-        if (simulatorInfo.initialValue !== undefined) {
+      if (simulatorInfo.initialValue !== undefined) {
```
TimGuiteDiamond: On line 510 you used this: `pvSimulator && pvSimulator.getValue()`
Why not use 
```js
simulatorInfo.initialValue && simulatorInfo.simulator && simulatorInfo.simulator.updateValue(simulatorInfo.initialValue);
```
or 
```js
if (simulatorInfo.initialValue && simulatorInfo.simulator)
{
    simulatorInfo.simulator.updateValue(simulatorInfo.initialValue);
}
```
here? Because you are especially interested in if they are *undefined* as opposed to just *falsy*?

willrogers:
I was still getting some updates. Here is the diff that fixed it:

```
diff --git a/src/connection/sim.ts b/src/connection/sim.ts
index 737ceb8..c7eebd2 100644
--- a/src/connection/sim.ts
+++ b/src/connection/sim.ts
@@ -75,7 +75,9 @@ abstract class SimPv {
   protected maybeSetInterval(callback: () => void): void {
     if (this.updateRate !== undefined) {
       setInterval((): void => {
-        callback();
+        if (this.subscribed) {
+          callback();
+        }
       }, this.updateRate);
     }
   }
@@ -89,10 +91,10 @@ class SinePv extends SimPv {
     updateRate?: number
   ) {
     super(pvName, onConnectionUpdate, onValueUpdate, updateRate);
-    setInterval((): void => {
+    this.maybeSetInterval((): void => {
       const value = this.getValue();
       this.onValueUpdate(this.pvName, partialise(value));
-    }, this.updateRate);
+    });
   }
 
   public getValue(): VType | undefined {
@@ -164,9 +166,8 @@ class SimEnumPv extends SimPv {
       isReadonly: true
     });
     this.onValueUpdate(this.pvName, partialise(this.getValue()));
-    setInterval(
-      (): void => this.onValueUpdate(this.pvName, partialise(this.getValue())),
-      this.updateRate
+    this.maybeSetInterval((): void =>
+      this.onValueUpdate(this.pvName, partialise(this.getValue()))
     );
   }
   public getValue(): VType {
@@ -201,9 +202,8 @@ class EnumPv extends SimPv {
       isConnected: true,
       isReadonly: false
     });
-    setInterval(
-      (): void => this.onValueUpdate(this.pvName, partialise(this.getValue())),
-      this.updateRate
+    this.maybeSetInterval((): void =>
+      this.onValueUpdate(this.pvName, partialise(this.getValue()))
     );
   }
```

I'm not sure if you want to update the tests as well. I'm not sure how you are handling the intervals in the tests.

talwrii:
> I'm not sure if you want to update the tests as well. I'm not sure how you are handling the intervals in the tests.

I only dealt with loc values. 

I don't like your patch :p (I'm going to change calls to onValueUpdate to calls to publish and keep onValueUpdate away from subclasses) - less duplication.


> I'm not sure if you want to update the tests as well. I'm not sure how you are handling the intervals in the tests.

I'm not (I was only dealing with local pv's which don't have setIntervals associated with them).

I shall.

src/connection/sim.ts
talwrii: Sorry kind of long winded...

TimGuiteDiamond: Thanks, Tom, for taking the time to explain this! I actually wasn't familiar with how the `&&` construct worked in JS before looking at your code and it can be easy to get carried away with a *cool new feature* so wanted to check when it is actually a relevant tool.

willrogers:
Also, Tim noticed that the `loc://` PVs are sending periodic updates. We only expect the `loc://` PVs to send an update when they change, much like real PVs.

Sim PVs have scheduled updates to simulate real PVs that do change periodically.

talwrii:
> Also, Tim noticed that the loc:// PVs are sending periodic updates. 

I wasn't able to reproduce this (both in a test and setting `log.setLevel(DEBUG)`).

talwrii:
Okay this should be ready for merge.

I know have tests for unsubscription and periodically updating sims.

willrogers:
You know what, I didn't see this when I was inspecting this branch yesterday. I think that's fixed on this branch already, sorry.

PR #99, opened 2019-11-11
Rule re render

willrogers:
@TimGuiteDiamond I believe that this should prevent the rule re-renders that you were seeing, but I would like you to confirm.


TimGuiteDiamond:
Yep this seems to do what it says on the tin, happy to merge

PR #98, opened 2019-11-06
Performance Testing

TimGuiteDiamond:
Addresses the goals of #94

Issue #94, opened 2019-11-01
Performance Testing and Improvements


TimGuiteDiamond:
We would like to understand how well this prototype performs when presented with a large number of PVs which are updating at a significantly high rate. This will also be of interest to many of the engineers who might use this type of program in the future.

[This paper](https://github.com/dls-controls/cs-web-proto/wiki/files/epics-ui-performance.pdf) provides a "standard" performance metric for comparing different EPICS GUI interfaces and is a useful starting point for future comparisons.

One prerequisite for performance testing is producing many PVs which can update at a significantly high rate. We can provide a first pass at this using the simulator.

Beyond this, we will need to use a database to generate actual PVs and connect to them. Our current Coniql plugin uses PVA, so we could set it up with an EPICS7 IOC with all these PVs and connect to that. It should also be possible to connect to that IOC from CS-Studio and Phoebus.

Tom Cobb will also provide advice and reference on previous GUI tests which were performed at Diamond and that information will be recorded here.

TimGuiteDiamond:
In addition to the above testing, some effort will be made to improve the efficiency of our React code. Most gains in this domain come from being able to reduce the number of times a component re-renders itself.

There are some strategies around this - [here](https://hackernoon.com/react-performance-primer-64fe623c4821) for example - however all optimisations have situations in which they work and situations in which they don't, and it is quite likely that our core program architecture could benefit from improvement before we begin chasing these gains. Nevertheless, it would be instructive to see where some of these optimisations can be useful and where they have a negligible or negative impact. This will be informed using the testing methods listed above.

TimGuiteDiamond:
Have been pointed towards using Python Soft IOCs to provide the requisite number of identical and similarly named PVs. Will also create a script to help me generate similar layouts of screens for CS-Studio, EDM and CS-Web-Proto.

TimGuiteDiamond:
Have written a script to perform this task, preliminary testing suggested that we have a lot of unnecessary re renders happening in the widgets which is massively slowing down performance, relative to CS-Studio.

This performance will be captured on a separate branch and recorded.

I will investigate a number of strategies to try and improve this performance:
- Use a simulation PV to remove potential slowdown as a result of ConiqlPlugin
- Use of React.memo and useMemo to test changes on inputs
- Custom memo checkers to expand embedded objects in the widget interface
- Flatten widget interface to allow prop checking to work better with a shallow check, thereby avoiding the need for custom checkers

[...]

PR #100, opened 2019-11-12
Ensure widgets re-render only if the right PV has changed.

willrogers:
@TimGuiteDiamond I believe this will catch the multiple render problem.

There's some refactoring to do here, but I'd like you to check that it does indeed fix the problem.


TimGuiteDiamond:
@willrogers have added some tests, as noted here and in #94 the by value tests do not actually appear to be necessary as the same object must be being used to represent this. I would like to add some further tests however related to changing minor elements of objects, such as alarm values and display ranges. Perhaps these will be forced to create entirely new objects which would therefore fail the shallow test.

I think the code should probably be cleaned up a bit, we might add or remove information from pvState arrays in the future, and perhaps a comment explaining why this works with respect to redux?

willrogers:
I refactored a bit. It's worth checking that the actual bug is still fixed!

willrogers:
Looks ok to me.

TimGuiteDiamond:
MultipleWidgetsOneRender
MultipleWidgetsOneRenderDevTools

Merged this into #98 and the above pictures show that indeed only the widgets which is subscribed to a PV value that **changes** is updated, the other two remain idel. Shown above with React profiler and Chrome devTools.

@willrogers this is ready to merge.

I will use this as a base to perform more testing with respect to an increased number of widgets on the screen.

Issue #96, opened 2019-11-01
Screen Layout


TimGuiteDiamond:
As we move towards having the capability to replicate the absolute layout of CS-Studio/EDM screens, we need to consider how we will actually do this and what it will look like. This will require some new components.

Points which I can think of:

- Match CS-Studio Synoptic/Embed/Detail layout
- How can we provide a mechanism to dictate which part of the screen displays should appear in from an action? (See CS-Studio's "Open OPI" action)
- What will this look like on mobile/flexible type layout? Could we have an "active" pane with options to activate a different pane? Should this inform what it looks like on desktop? Thinking of Malcolm here with its sliding panes for children/parent information
- Scientists often like having many panes open at the same time, how will we handle this?

I will add other points here as I think of them, anyone else feel free to add as well, but I think we can start moving towards this.

Sky's work in #88 is a good step in this direction! I think if this behaviour should be wrapped into an action and performed from the context menu for a shape, graphic or other widget and then merged as a good example.

aawdls:
This was superseded by #128.

Issue #75, opened 2019-10-25
Generic routing to screens


TimGuiteDiamond:
At the moment, our screens are all collected in App.tsx which has direct access to Page components which know exactly which JSON file to open.

It would be useful to add functionality to open any of the available JSON pages with given macroMaps and include this in the router context without having to explicitly create every page.

As an example use case, a button (or image) which opens another screen with a specified macro map, for use from the synoptic.

Most of the components are available for this, such as widgetFromJson, but it has not been implemented and may present some challenges.

This is an important and expected feature which we have not yet implemented.

skyfrench:
#88 includes a demo of dynamic routing. 
With this we can add e.g.
`<Link to={'dynamic/ionpExample/{"device":"SR03A-VA-IONP-01"}'}></Link>`
around a button (e.g. basic graphical button added in #85) and it'll open a new screen.

A couple of points for discussion:

- Where should the new screens open? Relates to #74. Do we want to divide screens into regions and have 'component' screens able to open in any of those. Or a new window/pop-up? 
- Should clicking to open a new screen (instead of a generic web page) actually be an action? Should it then be that only 'action buttons' can open screens, or any type of button? 
Should there be a base 'WidgetButton' with options to be (any or all of the following):
-- graphical (displaying as an image from file, or line, or circle)
-- action-full (the two existing actions plus open new screen) 
-- context menu'd?

TimGuiteDiamond:
Seems to work quite nicely, I guess moving towards the synoptic/embed/details layout is a detail that should be done later.
It really just depends where you instantiate the <Route> element

For the moment I think it would be nice to display this option via context menu on some kind of button, maybe add it to graphics later/for the demo?

TimGuiteDiamond:
We have achieved this to a good level for the time being with the DynamicPage component, and `useHistory` hook encoding the address, pages and macros in the HTML address.

@skyfrench if you agree this is working well then we can close this issue

PR #101, opened 2019-11-13
Layout & Routing

skyfrench:
Playing around with layouts and nested routing ...

(not proposing we merge this yet, opening request just for illustrative demo purposes).

#75 & #96


skyfrench:
@TimGuiteDiamond why is props.history undefined in ActionButtonWidget? 
(Simple Example -> DCM Button)

I'd like to pass the RouteComponentProps available and defined and usable in e.g. dynamicPage.tsx through to the widgets to pass into executeActions so that an action can use the browser router manually to open a route (manual equivalent of clicking 'Link'). But given how convoluted that sounds ... maybe I shouldn't be trying to do this?

skyfrench:
Couple of notes:
- Simplified App using #106.
-- but with modification: I have replaced 'name' with 'routePath' to allow completely arbitrary paths to be routed to a page
- There's now a:
-- top level json file (a 'synoptic') loaded by default in a div in App.tsx
-- second div which loads a json file which defines the page layout and possible route locations (e.g. left, right)
- Adding an action of type 'OPEN_PAGE' can be used to load any json page in any location with any macros, e.g. Component 1 (2) has an action to load 'ionpExample' in the 'left' location with a device macro for SR03A-VA-IONP-01 (SR04A-VA-IONP-01). The ionpExample now has a button with an action to open a further device screen in the 'right' location.

TimGuiteDiamond:
@skyfrench why does the ActionButton require a PV? Not all of the actions need them

skyfrench:
> @skyfrench why does the ActionButton require a PV? Not all of the actions need them

Good point - I don't think it necessarily does ...

TimGuiteDiamond:
It does take one in CS-Studio but not sure why :man_shrugging:

skyfrench:
> It does take one in CS-Studio but not sure why 🤷‍♂

It may be assumed all buttons have a status PV?

TimGuiteDiamond:
This has been taken forwards in #128 so perhaps could be closed here?

PR #104, opened 2019-11-13
Fixes

willrogers:
@TimGuiteDiamond @skyfrench a few small changes.

The most interesting is probably ab929ea, which rearranges `widgetDescriptionToComponent()` a bit to avoid a React error that we were seeing.

I allowed multiple rules, but really generic rules are another piece of work.


src/components/Widget/widget.tsx
```diff
@@ -192,23 +191,23 @@ export const PVWidget = (props: PVWidgetComponent): JSX.Element => {
...
-  const ruleProps = useRules(macroProps);
+  const ruleProps = useRules(macroProps) as PVWidgetComponent & { id: string };
   const [shortPvName, connected, readonly, latestValue] = useConnection(
     id,
     ruleProps.pvName
   );
-  let newProps: PVWidgetComponent & PvState = {
-    ...props,
+  let connectedProps = {
+    ...ruleProps,
```
TimGuiteDiamond: Yea I had been wondering why rule props didn't seem to be being used anywhere!

TimGuiteDiamond:
Looks good to me!

PR #105, opened 2019-11-14
Speed up PV Widgets

TimGuiteDiamond:


TimGuiteDiamond:
These changes were found not to produce any performance increases of note.

PR #106, opened 2019-11-14
Added DynamicPageWidget

TimGuiteDiamond:
Can load pages to itself based on it's name and the address parsing previously implemented. Plenty of work to do but a good start!

@skyfrench tried to implement this based on what you have already created. Quite clear to see how we could go from this to a synoptic/embed/details style layout although I am not completely clear on how to keep pages open in multiple boxes at the same time! Hoping you were working on this, based on what you said about using the router history.

### DIFF
--- a/public/flexiEmbedded.json
+++ b/public/flexiEmbedded.json
@@ -1,9 +1,7 @@
 {
   "type": "display",
   "containerStyling": {
-    "position": "absolute",
-    "x": "0px",
-    "y": "0px",
+    "position": "relative",
     "width": "100%",
     "height": "100%"
   },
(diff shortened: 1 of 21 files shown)


skyfrench:
> @skyfrench tried to implement this based on what you have already created. Quite clear to see how we could go from this to a synoptic/embed/details style layout although I am not completely clear on how to keep pages open in multiple boxes at the same time! Hoping you were working on this, based on what you said about using the router history.

I think the routes need to be nested to allow multiple boxes in view at the same time (see the 'Nested' link on #101).
e.g. Left renders paths of the form /:screen/:macros (using a Route with a non-exact match), right renders paths of form /:screen/:macros/:subscreen/:submacros
Then when navigating a Link to /:screen/:macros/:subscreen/:submacros from a button/action on the left screen, both the screen and the sub-screen stay.
I think this is the best that can be done using the router?

PR #108, opened 2019-11-15
Render from *.bob

TimGuiteDiamond:
Adds the capability to render *.bob files - check File Loading to view the example, from *simple.json* and *basic.bob*

Far from perfect but seems to work well for the time being, more components should be added, ProgressBar, SlideControl etc and the Display component should have a class attached to it which allows for overflow.

A slightly less easy piece of work which I am not particularly concerned with right now is converting colors from the bob xml style `red=128 green=128 blue=128` to CSS style `color=#888888`

Also should add a *groups* component to match up - basically is just a display with a header and border.


TimGuiteDiamond:
I have added actions to the bob rendering, only handling WRITE_PV at the moment. If we are happy with this approach then I think it is worth merging and we can continue this work at a later date.

TimGuiteDiamond:
@willrogers can we merge this?

willrogers:
I don't trust coveralls that much but it says the coverage has dropped quite a bit, and that makes sense because there are no tests for the grouping container or `fromBob.ts`.

TimGuiteDiamond:
Ok, will add some tests for them

TimGuiteDiamond:
I would like to add some further testing, particularly ensuring that the correct JSON is produced for steadily more complex bob/opi inputs but this increases the test coverage significantly above what it was so I will leave it for the time being.

TimGuiteDiamond:
Wasn't paying attention but found where I needed to make the change, this actually seems to work quite well now :) test coverage is up 2.5% compared to master, thanks to the additional tests for fromJson and fromBob so I am going to call that a win, give a quick review and then merge

src/components/FromBob/bobConversionUtils.ts
```diff
@@ -0,0 +1,242 @@
...
+
+import log from "loglevel";
+import convert from "xml-js";
+
+import { WidgetDescription } from "../Positioning/positioning";
+import { WidgetActions, WRITE_PV } from "../../widgetActions";
+
+interface BobDescription {
+  [key: string]: any;
+}
```
TimGuiteDiamond: Can we make this definition more useful? We know that the BobDescription will look like the compact JSON produced by the xml2js function, so could this be used to make it clearer what we expect to see?

src/components/FromBob/bobConversionUtils.ts
```diff
@@ -0,0 +1,242 @@
...
+
+export interface UnknownPropsObject {
```
TimGuiteDiamond: I think this is a necessary evil

src/components/FromBob/bobConversionUtils.ts
```diff
@@ -0,0 +1,242 @@
...
+      outputProps.visible = false;
+    }
+  } catch (e) {
+    log.error(
+      `Could not convert visible property ${inputProps.visible} to a number`
+    );
+  }
+};
+
+export const bobAvoidStyleProp = (
```
TimGuiteDiamond: Add a comment explaining why this seemingly pointless function is here

src/components/FromBob/bobConversionUtils.ts
```diff
@@ -0,0 +1,242 @@
...
+    write_pv: WRITE_PV, // eslint-disable-line @typescript-eslint/camelcase
+    WRITE_PV: WRITE_PV
+  };
+
+  // Turn into an array of Actions
+  let processedActions: WidgetActions = { executeAsOne: false, actions: [] };
+
+  actionsToProcess.forEach((action): void => {
+    log.debug(action);
+    let type: string = availableActions[action._attributes.type];
```
TimGuiteDiamond: Add an error handler here for the case where the action type is not found

Otherwise we will try and convert a slightly different opi or bob file and the whole thing will immediately crash

src/components/FromBob/bobConversionUtils.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  // Make sure widget is an array, otherwise it is seen as an object only
+  let widgetList = [];
+  if (Array.isArray(widget)) {
+    widgetList = widget;
+  } else {
+    widgetList = [widget];
+  }
+
+  // Check that the primary props were defined or use a default value
```
TimGuiteDiamond: Make a note about default values in bob files, this is a technical detail which will need implementation at a later date

src/components/FromBob/fromBob.ts
```diff
@@ -0,0 +1,154 @@
...
+      bobDescription = convertBobToWidgetDescription(
+        bob,
+        {
+          macros: bobMacrosToMacroMap,
+          background_color: bobBackgroundColor, // eslint-disable-line @typescript-eslint/camelcase
+          foreground_color: bobForegroundColor, // eslint-disable-line @typescript-eslint/camelcase
+          precision: bobPrecisionToNumber,
+          visible: bobVisibleToBoolen,
+          style: bobAvoidStyleProp,
+          actions: bobHandleActions
```
TimGuiteDiamond: Separate these into a separate object in the same way as widgetDict has been separated

src/components/GroupingContainer/groupingContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+  // This prevents the title being overwritten
+  // Could be changed or perhaps customisable as a prop
+  <fieldset
+    style={{
+      boxSizing: "border-box",
+      ...props.style
+    }}
+  >
+    <legend>{props.name}</legend>
+    <div style={{ position: "absolute", top: "20px", left: "20px" }}>
```
TimGuiteDiamond: How does this affect relatively positioned children inside this div? Perhaps margin could be used to provide the same effect?

src/components/Shape/shape.tsx
```diff
@@ -22,13 +21,12 @@ export const ShapeComponent = (
...
-        height: props.shapeHeight ? props.shapeHeight : "",
+        width: props.shapeWidth ? props.shapeWidth : "100%",
+        height: props.shapeHeight ? props.shapeHeight : "100%",
         borderRadius: props.shapeRadius ? props.shapeRadius : "",
-        transform: props.shapeTransform ? props.shapeTransform : "",
-        backgroundColor: props.shapeColour ? props.shapeColour : ""
+        transform: props.shapeTransform ? props.shapeTransform : ""
       }}
-    ></div>
+    />
```
TimGuiteDiamond: Fix this to still allow shape colour to come through

src/components/FromBob/bobConversionUtils.ts
TimGuiteDiamond: Have expanded this out a bit, could go on forever but perhaps not sensible to try and capture the entire bob schema within this interface?

src/components/FromBob/bobConversionUtils.ts
TimGuiteDiamond: Done

src/components/FromBob/bobConversionUtils.ts
TimGuiteDiamond: Done, just notes the error in the console and moves on as with the other conversion utils

src/components/FromBob/bobConversionUtils.ts
TimGuiteDiamond: Done

src/components/FromBob/fromBob.ts
TimGuiteDiamond: :+1:

src/components/GroupingContainer/groupingContainer.tsx
TimGuiteDiamond: Just making it a relatively placed div with 100% height and width seems to put it in a pretty good place (slightly different from xml but sensible), guess HTML people have already thought of this and sorted it for us!

src/components/Shape/shape.tsx
TimGuiteDiamond: Not going to do this right now

src/App.tsx
```diff
@@ -34,7 +34,7 @@ try {
   settings = {};
 }
 
-log.setLevel("warn");
+log.setLevel("debug");
```
willrogers: We don't want the default logging level to be `default`.

TimGuiteDiamond: My mistake, forgot to change that back!

PR #113, opened 2019-11-20
Restored styling capabilities

TimGuiteDiamond:
I have changed the code such that it behaves in the way I think it did before. In some places this has required defaulting heights and widths to 100%. This is acceptable because some components do not know whether they will be wrapped in another component which will give them size or not.

However, I have removed Tooltip from **Widget**, this should be added back in but only when tooltip is required on a **Widget**.


TimGuiteDiamond:
@willrogers amended the above changes to reinstate `TooltipWrapper` as part of a normal widget and to fix the tests :+1:

Issue #112, opened 2019-11-20
Flatten JSON description


TimGuiteDiamond:
In order to align the JSON file description more closely with the xml files produced by Phoebus/CS-Studio, we should flatten the description, removing the need specifically for the `containerStyling` and `widgetStyling` objects.

These objects were separated for use in the `recursiveWrapping` call which takes place inside `Widget` and `PVWidget`. The separation of props into these styling categories is still of use for our internal representation but not should not be required in this form for the JSON description.

This change would have the added bonus that xml -> JSON transformation would be more simple.

@willrogers has already made some progress on removing the *wrappers* object in #109 so I will build on that change.

Find below a JSON snippet of what this change should look like:

### Current JSON

```json
"type": "readback",
"containerStyling": {
       "position": "relative",
       "width": "100%"
},
"widgetStyling": {
       "fontWeight": "bold",
       "backgroundColor": "transparent"
},
"wrappers": {
      "alarmborder": true,
     "copywrapper": false
},
"wrappers
"pvName": "sim://ramp#1"
```

### Future JSON

```json
"type": "readback",
"position": "relative",
"width": "100%",
"fontWeight": "bold",
"backgroundColor": "transparent",
"tooltip": "Custom tooltip:\n${pvName}\n${pvValue}",
"alarmBorder": true,
"pvName": "sim://ramp#1"
```

Some of the labels which we give to these values are sure to change as we move forwards but the overall effect can already be seen.

PR #114, opened 2019-11-20
Flat json

TimGuiteDiamond:
Start of work on #112 

Demonstrated with frontPage.json

Works alongside current layout.

Will move on to flattening the wrapper props.


TimGuiteDiamond:
@willrogers combined with the work on #109 this is pretty lean now

See frontPage.json for the flat JSON style

willrogers:
Yes, looks fine. There is a lot of the old-style configuration left around in json files though.

TimGuiteDiamond:
Agreed but I didn't think it was too important to change this now. Plenty of the files can probably be ignored when we are setting up our final example anyway.

PR #116, opened 2019-11-22
Actions

willrogers:
@TimGuiteDiamond another minor refactor.


TimGuiteDiamond:
Can we merge this? Doesn't seem like a drastic change and I think the more explicit Actions, along with separation between redux actions and "other" actions is a good thing :+1:

PR #117, opened 2019-11-26
Dynamic Close Button

TimGuiteDiamond:
Add a close button to Dynamic Page Widget so that they can be easily closed

### DIFF
--- a/public/splitPage.json
+++ b/public/splitPage.json
@@ -12,29 +12,14 @@
       "position": "relative",
       "children": [
         {
-          "type": "flexcontainer",
+          "type": "dynamicpage",
           "position": "relative",
-          "width": "auto",
-          "flexflow": "column",
-          "children": [
-            {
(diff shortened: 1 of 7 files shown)


TimGuiteDiamond:
@skyfrench added a close button which renders as part of the route rendering, currently just opens a page with macro set to "CLOSED"

Could you please add a CLOSE_PAGE action (or similar name) which takes a location and removes that locations information from the path

i.e if the current page is http://localhost:3000/left/ionpExample/%7B%22device%22:%22SR04A-VA-IONP-01%22%7D/right/ionpExample/%7B%22device%22:%22SR04A-VA-IONP-01%22%7D/

and I pass {type: CLOSE_PAGE, location: "left"}

the new address should be http://localhost:3000/right/ionpExample/%7B%22device%22:%22SR04A-VA-IONP-01%22%7D/

skyfrench:
Added to #101.

PR #119, opened 2019-11-26
Include arrays when resolving macros in props.

### LINKED ISSUE #118: Handle Macros in Actions
Actions are stored inside an Actions array, which is not handled by the `useMacros` hook.

A naive implementation of checking arrays with Array.isArray leads to infinite recursion as JS objects have many hidden arrays which go on forever.

I have hardcoded a solution for actions in #101 but this should be generalised to other arrays that we want to check.

willrogers:
Fixes #118.

@TimGuiteDiamond

### DIFF
--- a/public/examplePage.json
+++ b/public/examplePage.json
@@ -193,24 +193,27 @@
         "height": "",
         "padding": ""
       },
+      "macroMap": {
+        "prefix": "sim://"
+      },
       "actions": {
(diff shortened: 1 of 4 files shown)


TimGuiteDiamond:
Yep, seems to work well :+1: 

Presume the children thing is to stop us having to solve these macros multiple times?

Happy to merge!

PR #125, opened 2019-12-03
Optional pvname

willrogers:
@TimGuiteDiamond make PV name optional on PV widgets.

It's often the case that you don't want a PV.

### DIFF
--- a/src/components/Widget/widget.tsx
+++ b/src/components/Widget/widget.tsx
@@ -131,7 +131,7 @@ type WidgetComponent = WidgetProps & { baseWidget: React.FC<any> };
 
 // Internal prop types object for properties which are not in a standard widget
 const PVExtras = {
-  pvName: PropTypes.string.isRequired,
+  pvName: PropTypes.string,
   alarmBorder: PropTypes.bool
 };
 // PropTypes object for a PV widget which can be expanded
--- a/src/hooks/useConnection.test.ts
+++ b/src/hooks/useConnection.test.ts
@@ -1,0 +1,42 @@
+import { useSelector } from "react-redux";
+import { vdouble } from "../vtypes/vtypes";
+import { useConnection } from "./useConnection";
+
+const TEST_VAL = vdouble(2);
+
+// Mock useSubscription.
+jest.mock("./useSubscription", (): object => {
+  return {
+    useSubscription: jest.fn()
+  };
+});
+// Mock useSelector.
+jest.mock("react-redux", (): object => {
+  return {
+    useSelector: jest.fn()
+  };
+});
+// This has to be done in a second step because Jest does the
+// mocking before we have access to other imports (vdouble).
+(useSelector as jest.Mock).mockImplementation((pvName: string): any => {
+  return {
+    PV1: [{ value: TEST_VAL, connected: true, readonly: false }, "PV1"]
+  };
+});
+
+describe("useConnection", (): void => {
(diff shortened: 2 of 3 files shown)


TimGuiteDiamond:
What are the use cases for this? Our separation of widget and PV widget is based on the assumption that these things need to be treated separately, if that is not the case then perhaps they can be merged?

willrogers:
That's not the worst idea. It would mean that every widget (even e.g. `Shape` and `Line`) has `pvName` and `alarmBorder` props, which is not the same as CS-Studio or Phoebus. In return it would simplify the code.

All widgets would also have a default tooltip of `${pvName}/${pvValue}`, although we could fix that.

TimGuiteDiamond:
> It's often the case that you don't want a PV.

It would potentially simplify the code, but could you expand on this statement for me? Perhaps there are come examples in this project you could point to?

willrogers:
In cs-web-proto `ActionButton` has a PV but doesn't necessarily need one.

I just checked a directory of opi files and the following widgets are PV widgets but didn't have a PV name:

```
org.csstudio.opibuilder.widgets.ActionButton
org.csstudio.opibuilder.widgets.Ellipse
org.csstudio.opibuilder.widgets.MenuButton
org.csstudio.opibuilder.widgets.Rectangle
org.csstudio.opibuilder.widgets.polygon
```

But there you don't see any consistency between what is a Widget (e.g. Label) and what is a PVWidget (e.g. Rectangle). So how useful is it to have non-pv-widgets? I'm not so sure.

TimGuiteDiamond:
In CS-Studio the label cannot take a pv_name but rectangles can. This appears to be mainly to allow rectangles to use the alarm border.

I'm not sure how often this is used. It might be possible to refactor widgets to all be PVWidgets, but only some of them require a PVName, not sure how we would handle this through propTypes and so on though. Certainly a different set of default wrappers could be applied depending on whether or not the pvName parameter was provided, and then overridden by specific settings on the widget, the alarmBorder prop for example.

willrogers:
Odd that a rectangle can have an alarm border but a label can't, no?

TimGuiteDiamond:
I don't disagree with you, perhaps things are different on Phoebus? Anyway, I think that suggests we have some wiggle room to decide what we want to do.

I think I would support merging PVWidget and non-PV widget, provided we could find a way to alert when a widget that should definitely have a PV such as a readback is missing it in its props.

At its most simple level, the issue you raised could be amended by having a button that is not a PVWidget, I'm not even sure it should have a PV to be honest.

Following our conversation just now I don't see any problems with this, but some tests should be added to ensure things are working properly. For example, a widget with pvName: "test" and alarmborder: false should not have an alarmborder. At the moment the wrapper is quite simple because we just push things on as the props are tested, but perhaps this could get a little more complex if we are dynamically adding/removing things? Guess it may end up just being a succession of conditional tests.

skyfrench:
I think you could want a button which has a PV - e.g. a button to represent a status PV of some component. You could also want an image (line, shape, picture) with a status PV - but similarly also want one without.

As you comment there will be certain types of widget which don't make sense without a PV (e.g. readback), but others for which it's optional.

willrogers:
But then there's not really such a thing as a mandatory string, you can just give it an empty one.

I think this is pretty harmless, so I'll merge it, but we'll need to do more thinking about this when we come around to re-thinking prop types.

PR #126, opened 2019-12-09
Extract execute as one property from bob files

TimGuiteDiamond:
Fix for bob files only. Opi files have `hook` attribute for executing first action in list on mouse click and `hook_all` for executing all on mouse click. Not going to handle that now.


TimGuiteDiamond:
Note that this change doesn't really have a material effect on the demo as we are using a rectangle which doesn't respond to mouse clicks anyway.

PR #127, opened 2019-12-10
Add implementation of widget ADFFViewer

EmilioPeJu:
It allows rendering a video stream based on an Area Detector FFmpeg plugin PV prefix.

### DIFF
--- a/public/frontPage.json
+++ b/public/frontPage.json
@@ -70,6 +70,20 @@
           "pvName": "loc://pv2"
         },
         {
+          "type": "input",
+          "position": "relative",
+          "width": "45%",
+          "margin": "2%",
+          "pvName": "loc://pv-prefix-1:MJPG_URL_RBV"
+        },
+        {
+          "type": "readback",
+          "position": "relative",
+          "width": "45%",
+          "margin": "2%",
+          "pvName": "loc://pv-prefix-1:MJPG_URL_RBV"
+        },
+        {
           "type": "readback",
           "position": "relative",
           "width": "45%",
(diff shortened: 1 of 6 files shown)


TimGuiteDiamond:
Thanks Emilio, this looks very nice and is quite small and clean :+1: 

Can I ask why you provided the `prefix` prop rather than just using pvName as is? Is this how the current implementation works?

hir12111:
It's a way of organizing elements in a tree, area detector plugins use a prefix for all its PVs, if we pass the prefix instead of the pvName that will allow the widget to be expanded easily (the same widget could be using many different PVs)... so I assumed there will be a area detector hierarchy with a top level area detector widget passing the prefix to each one of the plugins widgets, I think it's cleaner letting each plugin's widget deal with all its related PVs.

TimGuiteDiamond:
@willrogers this widget is fairly simple but looks to me like a subset of a potential `Image` component. Should I work on the more general component and make this fit into that or work with Emilio to clean up this component to merge it?

willrogers:
Good point, an image component wouldn't look much different. We should make a generic image component then see whether we need any extension to handle ffmpeg.

First priority is getting changes from the demo merged and deciding what the default app looks like,

hir12111:
It would be easy to pull an image widget from the RawADFFComponent, however I would still keep this widget as it is, decoupled from an image widget, the reason is that an image could have a huge number of transformations and CSS properties without affecting performance, if we use an image widget for the ADFFViewer, then every change in CSS properties or any transformation we do in that widget will be applied to every single frame of the stream... making performance much worse,  thus it's nice keeping them separate and having RawADFFComponent with a minimum number for transformations (just re-sizing  to the parent container)

aawdls:
Closing as we now have a widget for camera streams in cs-web-lib.

PR #128, opened 2019-12-12
Synoptic

TimGuiteDiamond:
Attempt to create an approximation of the synoptic layout as used in CS-Studio and EDM, along with some other demonstrations such as a mobile friendly layout and performance page


src/widgetActions.ts
```diff
@@ -41,14 +58,74 @@ export const getActionDescription = (action: WidgetAction): string => {
...
+        openPage(action, history);
+      } else {
+        log.error("Tried to open a page but no history object passed");
+      }
+      break;
+    case CLOSE_PAGE:
+      if (history) {
+        closePage(action, history);
+      } else {
+        log.error("Tried to open a page but no history object passed");
```
willrogers: Close not open...

willrogers:
Quite a few changes here but most of the important ones are quite likely to be revisited, so I'm happy to merge. Unfortunately Travis seems a bit busy right now.

TimGuiteDiamond:
Yea I'm quite happy to fix some small changes if you have any more suggestions, otherwise we should probably make issues for the bigger issues and move on.

Would be nice to wait for Travis to verify this.

PR #132, opened 2019-12-13
Image Widget

TimGuiteDiamond:
First pass at a generic image widget


TimGuiteDiamond:
Note that the CS-Studio image widget does not expect a pv name, so I have matched that capability

PR #134, opened 2019-12-17
Updates2

willrogers:
@TimGuiteDiamond updates to lots of packages including TypeScript 3.7.

There are a few fixes and workarounds to enable this, but one is a bit clunky: the rearranging of widgetActions.

I'll merge it on the basis that it will probably be refactored again, but there might well be a better way to do this. See also https://stackoverflow.com/questions/59373461/proptypes-oneoftype-not-working-as-i-expect-with-typescript


willrogers:
Once this one is merged I will do the file rearrangements.

TimGuiteDiamond:
Is this a symptom of the "Prop types are not quite runtime versions of TS types" issue? The original layout would work fine in typescript wouldn't it? And seems a bit nicer to me.

Issue #140, opened 2020-01-27
Popover causing unnecessary renders


TimGuiteDiamond:
Popover is being used by the tooltip component to provide information when a widget is middle clicked on.

However, it is causing a clear performance hit. We can see below the profile for 5 readback widgets connected to the same PV. On the right hand side, we can see the widget render tree updating under one tree reconciliation, as expected. On the left hand side however, we see multiple reconciliations each just for one popover. Clearly as the number of widgets increases, this will be increasingly detrimental to performance.

![popover causing issues](https://user-images.githubusercontent.com/50404435/73168650-25e13180-40f2-11ea-8759-f0b35dcb426e.png)

By commenting out the following lines in widget.tsx, we remove this performance detriment as demonstrated in the below image:
```typescript
const resolvedTooltip = resolveTooltip(connectedProps);
connectedProps.resolvedTooltip = resolvedTooltip;
```

![no popover, no problem](https://user-images.githubusercontent.com/50404435/73168921-d51e0880-40f2-11ea-86ea-ffe61c061487.png)

This warrants some investigation into why this is occurring and how we can prevent these unnecessary tree reconciliations, and implement our popover functionality efficiently.

TimGuiteDiamond:
Fixed in #141 with small changes to popover implementation

![popover fixed](https://user-images.githubusercontent.com/50404435/73172994-34cce180-40fc-11ea-9c98-695504f62c95.png)

PR #141, opened 2020-01-27
Popover fix

TimGuiteDiamond:
Fixes issue in #140


TimGuiteDiamond:
@willrogers significant speedup confirmed, should be ready to merge

PR #142, opened 2020-01-27
Performance changes

TimGuiteDiamond:


src/connection/sim.ts
```diff
@@ -20,11 +20,14 @@ import { alarm, ALARM_NONE } from "../types/vtypes/alarm";
...
-function partialise(value: VType | undefined): PartialVType | undefined {
+function partialise(
+  value: VType | undefined,
+  type?: string
+): PartialVType | undefined {
   if (value === undefined) {
     return undefined;
   } else {
-    return vtypeInfo(value, {});
+    return vtypeInfo(value, { type: type });
```
TimGuiteDiamond: @willrogers this change allows us to hardcode in type values for the simulator values, see the rest of this file for the values I provided. This means the type string is consistent across development and production builds for the simulator, otherwise it defaults to `original.constructor.name` which is abbreviated in the production build.

PR #143, opened 2020-01-31
Move connection hook to improve performance

TimGuiteDiamond:
The connection hook is required at runtime, but many of the other hooks (useMacros etc) are not. Therefore, this PR moves the connection hook into a separate widget which is rendered after the load-time hooks in order to separate concerns. This appears to have provided an approximately 36% increase in speed over what is currently on the master branch when measuring both with production build and the profile flag (`npm run build -- --profile`)

It also does not disrupt our current widget layout but does give us a method for concerning load-time and runtime concerns.


TimGuiteDiamond:
This PR currently prevents us from calculating rules at runtime, but we can't do this effectively anyway for many props (such as position, font etc). I think this should be looked at as a separate concern.

src/ui/widgets/widget.tsx
```diff
@@ -151,6 +151,41 @@ const recursiveWrapping = (
   }
 };
 
+const WrappedComponents = (props: {
+  components: React.FC<any>[];
+  containerStyling: object;
+  widgetStyling: WidgetStyling | null;
+  containerProps: any & { id: string };
+  widgetProps: any;
```
TimGuiteDiamond: I would like to stop these allowing `any` through in the future

TimGuiteDiamond:
image
Time in ms for rendering 100 readbacks widgets, measured 10 times at 1 Hz to allow plenty of time for render to finish before new value change

All readbacks looking at the same PV. Clear speedup from hooked and wrapped components changes. Performance comparable to dbwr for this test.

willrogers:
Interesting that this branch improves performance, as it seems a fairly minor change. As the performance is comparable to what you called 'hooked-components' I think this one is more useful.

PR #145, opened 2020-02-05
Resolve macros on the props object rather than creating a new one.

willrogers:
Creating a new props object would lose any methods if the object was an
instance of a class.


TimGuiteDiamond:
What are examples of times we would expect a class instance to be passed through `useMacros`?

willrogers:
I want to pass `Color` and `Font` objects, for example. I realised when passing a `Color` object that it wasn't one by the time it got to rendering.

TimGuiteDiamond:
:+1:

PR #146, opened 2020-02-05
Remove the widgetStyling prop.

willrogers:
All such props are now required at the top level of the widget itself.

@TimGuiteDiamond this is what this change would look like. I am tempted to merge this and move on. I might revert some of this later depending on how it pans out.

### DIFF
--- a/src/ui/widgets/ActionButton/actionButton.tsx
+++ b/src/ui/widgets/ActionButton/actionButton.tsx
@@ -1,6 +1,6 @@
 import React, { useContext } from "react";
 import { WidgetActions, executeActions } from "../widgetActions";
-import { PVComponent, PVWidget, PVWidgetPropType } from "../widget";
+import { PVWidget, PVWidgetPropType } from "../widget";
 import classes from "./actionButton.module.css";
 import { useHistory } from "react-router-dom";
 import { registerWidget } from "../register";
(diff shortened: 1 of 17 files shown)


src/ui/widgets/ActionButton/actionButton.tsx
```diff
@@ -32,7 +33,12 @@ export const ActionButtonComponent = (
     <button
       className={classes.actionbutton}
       onClick={props.onClick}
-      style={props.style}
+      style={{
```
TimGuiteDiamond: Should `color=props.foregroundColor` be included here?

src/ui/widgets/FlexContainer/flexContainer.tsx
```diff
@@ -11,18 +11,14 @@ const FlexProps = {
...
   const { flexFlow = null } = props;
   if (flexFlow !== null) {
     classNames.push(classes[flexFlow]);
   }
-  return (
-    <div className={classNames.join(" ")} style={props.style}>
-      {props.children}
-    </div>
-  );
+  return <div className={classNames.join(" ")}>{props.children}</div>;
```
TimGuiteDiamond: I think FlexContainer needs to know how big it should be? Is this still working as expected?

src/ui/widgets/GroupingContainer/groupingContainer.tsx
```diff
@@ -11,15 +11,14 @@ const GroupingContainerProps = {
...
+  props: InferWidgetProps<typeof GroupingContainerProps>
 ): JSX.Element => (
   // Uses an inner margin for children similar to Phoebus
   // This prevents the title being overwritten
   // Could be changed or perhaps customisable as a prop
   <fieldset
     style={{
-      boxSizing: "border-box",
-      ...props.style
+      boxSizing: "border-box"
```
TimGuiteDiamond: This might also want to know its size?

willrogers:
@TimGuiteDiamond yes, there was some handling of the position here that might not still be there. How did it work?

PR #147, opened 2020-02-07
Rules

willrogers:
@TimGuiteDiamond these changes make the rules a lot more capable and makes things a bit nicer to work with. It also improves the OPI loading quite a bit.

It's another big refactor, so I guess it's an example where easy-to-run performance metrics would be nice - there could easily be a performance impact but it's not straightforward to know.

Anyway, I don't think it's necessarily ready to merge but I think you should have a look.


willrogers:
Oh, I've introduced a dependency cycle so we'll need to fix that.

TimGuiteDiamond:
Having only had a quick glance this looks good!

When I tried pulling this branch, it popped up with some work @skyfrench did on a branch also called `rules` in October. One of the changes she had made was to use the `evaluate` function from the `mathjs` package. Is this of use to us now? `mathjs` is still in our `package.json`

src/types/props.ts
```diff
@@ -0,0 +1,35 @@
...
+  | MacroMap
+  | WidgetActions;
+
+interface Expression {
+  boolExp: string;
+  value: string;
+  convertedValue?: GenericProp;
+}
+
+interface PV {
```
TimGuiteDiamond: Could we give this a more descriptive name? It appears to be only used for `Rules` below but could still cause confusion later

src/ui/widgets/EmbeddedDisplay/jsonUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+function jsonParseBorder(name: string, jsonBorder: any): Border {
+  const styles: { [key: string]: BorderStyle } = {
+    none: BorderStyle.None,
+    line: BorderStyle.Line,
+    dashed: BorderStyle.Dashed,
+    dotted: BorderStyle.Dotted,
+    groupbox: BorderStyle.GroupBox
+  };
+  return new Border(
+    styles[jsonBorder.style?.toLowerCase()],
```
TimGuiteDiamond: This is optionally chained, but style is required for the Border type. What style will you get if none is provided? Should this raise an error?

src/ui/widgets/EmbeddedDisplay/jsonUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+  height: string;
+  width: string;
+  children: JsonDescription[];
+  [key: string]: any;
+}
+function jsonParseColor(name: string, jsonColor: string): Color {
+  return Color.parse(jsonColor);
+}
+
+function jsonParseBorder(name: string, jsonBorder: any): Border {
```
TimGuiteDiamond: Could `jsonBorder` have a more specific type?

src/ui/widgets/EmbeddedDisplay/jsonUtils.ts
```diff
@@ -0,0 +1,97 @@
...
+    bold: FontStyle.Bold,
+    "bold italic": FontStyle.BoldItalic
+  };
+  return new Font(
+    styles[jsonFont.style?.toLowerCase()],
+    jsonFont.size,
+    jsonFont.typeface
+  );
+}
+
```
TimGuiteDiamond: Same as above - perhaps there is a reason for this approach?

src/ui/widgets/FlexContainer/flexContainer.tsx
```diff
@@ -1,13 +1,20 @@
...
+  InferWidgetProps,
+  BorderPropOpt
+} from "../propTypes";
 
 const FlexProps = {
   flexFlow: ChoicePropOpt(["rowWrap", "column", "row", "columnWrap"]),
-  children: ChildrenPropOpt
+  children: ChildrenPropOpt,
+  border: BorderPropOpt
 };
```
TimGuiteDiamond: Should containers such as this have background colour options?

src/ui/widgets/Input/input.tsx
```diff
@@ -15,7 +17,7 @@ export interface InputProps {
   onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
   onBlur: (event: React.ChangeEvent<HTMLInputElement>) => void;
   onClick: (event: React.MouseEvent<HTMLInputElement>) => void;
-  style?: object;
+  font?: Font;
 }
```
TimGuiteDiamond: No colour props for `Input`?

public/json/frontPage.json
```diff
@@ -53,188 +66,46 @@
           "position": "relative",
           "width": "45%",
           "margin": "2%",
-          "pvName": "loc://pv1",
```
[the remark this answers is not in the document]
willrogers: Sorry, this one was changed for testing and I erroneously included the changes.

willrogers:
@TimGuiteDiamond the answer to several of these queries is that this puts the framework in place so that it's easy to add things e.g. more border styles, colours for the input widget. I don't feel the need to make all these things comprehensive yet.

src/types/props.ts
willrogers: Makes sense.

src/ui/widgets/EmbeddedDisplay/jsonUtils.ts
willrogers: Good question, this was probably just a shortcut I took.

src/ui/widgets/EmbeddedDisplay/jsonUtils.ts
willrogers: Probably.

src/types/color.test.ts
```diff
@@ -0,0 +1,25 @@
+import { Color } from "./color";
+
+describe("Color", (): void => {
+  // This test fails because we can't create elements in testing.
+  it.each<[string, Color]>([
+    ["green", Color.BLACK],
+    ["red", Color.BLACK]
+  ])("fails to parse color names correctly", (name, expectedColor): void => {
+    expect(Color.parse(name)).toEqual(expectedColor);
```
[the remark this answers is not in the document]
willrogers: The mechanism for figuring out the color codes requires creating an element; we can't do this in the tests by default. There may well be a way to do it in testing, but for now this test captures the actual behaviour. I'll expand the comment and maybe flag as an expected failure [edit no I can't find a way to do that].

public/json/beamline.json
```diff
@@ -24,7 +36,9 @@
               "width": "90%",
               "height": "20px",
               "textAlign": "center",
-              "fontWeight": "bold",
+              "font": {
+                "style": "bold"
+              },
```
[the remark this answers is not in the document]
willrogers: Ideally prop-types would flag these things.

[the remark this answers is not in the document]
TimGuiteDiamond: We should make a note of this kind of thing as an example of what prop types does not help us solve, so that we could compare it to another object -> type package in the future

TimGuiteDiamond:
I've resolved all the conversations which seem like they fall under the above comment, if the rest are resolved then it should be fine to merge.

This would also be a good opportunity to try and follow the performance measuring techniques I have outlined in the wiki, to provide feedback on both the instructions and the overall process.

TimGuiteDiamond:
Cool, looks good to me and the opi/bob conversion is a pretty big improvement :)

Issue #149, opened 2020-02-12
Performance Measurement


TimGuiteDiamond:
We have made some good headway regarding performance and measuring this at different levels.

However, the current methodology is manual and time consuming. It would be great to have an automated measurement tool of some kind. Ideally this would integrate with our CI platform to make it obvious to us when we have introduced significant performance slowdown into the application.

As a first step it would be useful to have a tool which measures renders at various levels and outputs this data in a useful format, to speed up the process of measuring multiple renders and extracting the render time, reconciliation time etc.

[React Profiler](https://reactjs.org/docs/profiler.html) looks like a useful tool and [this article](https://medium.com/life-at-paperless/how-to-use-the-react-profiler-component-to-measure-performance-improvements-from-hooks-d43b7092d7a8) describes its use and benefit, as well as linking to a live project on Github which is using this profiling mechanism.

TimGuiteDiamond:
The Profiler component has indeed proven successful and useful.

It is amuch easier way of extracting the information we are interested in and should allow us to perform useful comparisons on branch changes in the future. Although the method I lay out below is manual, it has the potential to be automated much more in the future.

# Verifying previous results

We know from our previous measurements that the `master` branch is currently quite good at rendering many readbacks which are all looking at a single PV, and that without throttling, performance takes a big hit when looking at many different PVs.

To verify the usefulness of the Profiler component, I performed the same experiments again, measuring 3 statistics available through the [`onRender` callback](https://reactjs.org/docs/profiler.html#onrender-callback):

- actualDuration - time spent rendering the committed update
- baseDuration - estimated worst case cost of rendering the subtree
- reconciliationTime - commitTime - startTime (called React Tree Reconciliation in the Chrome Devtools performance view)

# Methodology

I initially set up the Profiler component around the top level DynamicPage widget, which holds the performance page and all of the Readbacks when they are shown. For our `master` branch, a simple use of the `onRender` callback to provide comma separated outputs sufficed. I tested the performance page with various numbers of Readbacks, all looking at `sim://ramp#00`.

[...]

PR #150, opened 2020-02-13
Performace measurement

TimGuiteDiamond:
Use the React profiler component to produce highly useable information regarding render times for the main subtree.

Would be nice to wrap this up into its own file but at the moment it is dependent on the simulation time setting. I think this should perhaps be moved into `settings.ts` anyway as in the future other developers or sites might want to tweak this setting without having to edit and rebuild the code.

It would also be interesting to see how we can make this output more useful and generate it by running a test, either on our own machines or as part of the CI job.

### DIFF
--- a/.lighthouserc.json
+++ b/.lighthouserc.json
@@ -1,0 +1,11 @@
+{
+  "ci": {
+    "collect": {
+      "startServerCommand": "serve -s build -l 3000",
+      "url": [
+        "http://localhost:3000/app/performance/{}/",
+        "http://localhost:3000/app/performanceDifferentPVs/{}/"
+      ]
+    }
+  }
+}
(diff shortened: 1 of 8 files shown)


TimGuiteDiamond:
More info about this branch's use has been collected here: #149

TimGuiteDiamond:
@willrogers have made some changes which make this more merge-ready. Also noticed a file `src/s.ts` was added during your `restructure` PR a while back, is that meant to be there? Appears to do the same thing as `src/settings.ts` ?

willrogers:
Looks nice, thanks.

willrogers:
I don't know why but I am getting a test failure on my machine that we don't see on Travis.

willrogers:
`npm install` fixed it. It's a bit odd but never mind.

PR #154, opened 2020-02-21
Throttle middleware

willrogers:
This is a nicer way of implementing the throttling.

I won't merge it yet as I should write some tests.

### DIFF
--- a/src/redux/csState.ts
+++ b/src/redux/csState.ts
@@ -75,18 +75,7 @@ export function csReducer(state = initialState, action: Action): CsState {
   switch (action.type) {
     case VALUE_CHANGED: {
       const newValueCache: ValueCache = { ...state.valueCache };
-      const { pvName, value } = action.payload;
-      const pvState = state.valueCache[pvName];
-      let newValue: VType | undefined;
-      if (value instanceof VType) {
-        newValue = value;
-      } else {
-        newValue = mergeVtype(pvState.value, value);
-      }
-      const newPvState = Object.assign({}, pvState, {
-        value: newValue
-      });
-      newValueCache[action.payload.pvName] = newPvState;
+      updateValueCache(state.valueCache, newValueCache, action);
       return { ...state, valueCache: newValueCache };
     }
     case VALUES_CHANGED: {
--- a/src/redux/csState.ts
+++ b/src/redux/csState.ts
@@ -7,7 +7,8 @@ import {
   WRITE_PV,
   CONNECTION_CHANGED,
   MACRO_UPDATED,
-  UNSUBSCRIBE
+  UNSUBSCRIBE,
+  ValueChanged
 } from "./actions";
 import { VType } from "../types/vtypes/vtypes";
 import { mergeVtype } from "../types/vtypes/merge";
@@ -53,7 +54,7 @@ export interface CsState {
 function updateValueCache(
   oldValueCache: ValueCache,
   newValueCache: ValueCache,
-  action: any
+  action: ValueChanged
 ): void {
   const { pvName, value } = action.payload;
   const pvState = oldValueCache[pvName];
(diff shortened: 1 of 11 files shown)


src/redux/csState.ts
```diff
@@ -49,6 +50,25 @@ export interface CsState {
...
+  } else {
+    newValue = mergeVtype(pvState.value, value);
+  }
+  const newPvState = Object.assign({}, pvState, {
+    value: newValue
+  });
+  newValueCache[action.payload.pvName] = newPvState;
+}
+
 export function csReducer(state = initialState, action: Action): CsState {
```
TimGuiteDiamond: What does `state = initialState` do?

src/redux/csState.ts
```diff
@@ -68,6 +88,13 @@ export function csReducer(state = initialState, action: Action): CsState {
       newValueCache[action.payload.pvName] = newPvState;
```
TimGuiteDiamond: Could this use `updateValueCache` as well?

src/redux/csState.ts
```diff
@@ -49,6 +50,25 @@ export interface CsState {
   subscriptions: Subscriptions;
 }
 
+function updateValueCache(
+  oldValueCache: ValueCache,
+  newValueCache: ValueCache,
+  action: any
```
TimGuiteDiamond: As there are only a number of actions which will be useful for this function, perhaps the type could be specified further? I suppose it won't matter much as it will mainly be used at runtime but it might make the code easier to reason about.

src/redux/throttleMiddleware.ts
```diff
@@ -0,0 +1,48 @@
...
+    store.dispatch({ type: VALUES_CHANGED, payload: [...this.queue] });
+    this.queue.length = 0;
+    this.ready = false;
+  }
+}
+
+export const throttleMiddleware = (updater: UpdateThrottle) => (store: any) => (
+  next: any
+) => (action: any): any => {
+  if (action.type === VALUE_CHANGED) {
```
TimGuiteDiamond: Should this also be able to handle `VALUES_CHANGED`? Perhaps in the future the server or complex widgets could provide bundled value updates which would want to use this mechanism?

src/redux/csState.ts
willrogers: This is the way you initialise the store.

src/redux/csState.ts
willrogers: Yes, this was a lazy `any`.

src/redux/csState.ts
willrogers: Yep.

src/redux/throttleMiddleware.ts
willrogers: Yes, that sounds very possible. I will leave it as it is, because if we do want such a feature it will be possible to add it at the time and it's best to do it when you know more clearly what you need.

TimGuiteDiamond: Ok, I guess in terms of complex widgets we would need to write a custom dispatcher and therefore a new hook anyway?

src/redux/csState.ts
TimGuiteDiamond: Just wondering about the syntax - why is it not just `csReducer(initialState...`)? Does it take the value only the first time it is called or something like that? Or is it because initialState is the global state value we are trying to protect?

TimGuiteDiamond:
Making the period configurable from the settings is definitely a good idea :+1: 

The settings file could get out of hand at some point though :sweat_smile:

TimGuiteDiamond:
I'd like to do a quick profiler check on this before we merge it

willrogers:
Let's add a description of what a profiler check entails to the wiki page.

TimGuiteDiamond:
Performance appears to match the previous throttle set up we had, so this can be merged :+1:

PR #155, opened 2020-02-26
Waveforms

TimGuiteDiamond:
Added simulated waveforms and performance page to especially test them

### DIFF
--- a/public/json/performanceWaveforms.json
+++ b/public/json/performanceWaveforms.json
@@ -1,0 +1,1215 @@
+{
+  "type": "display",
+  "containerStyling": {
+    "position": "absolute",
+    "x": "0px",
+    "y": "0px",
+    "width": "1300px",
+    "height": "500px"
+  },
+  "widgetStyling": {
+    "backgroundColor": "#999999"
+  },
+  "children": [
+    {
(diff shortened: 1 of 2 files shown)


TimGuiteDiamond:
Results from profiling the waveforms against standard ramp PV, with 100 PVs

Waveforms with size of 1: a6323fd
Waveforms with size of 10: 7456a55

## Development Build

| Application Setup | Size of waveform (elements) | Render Time (ms) |
| --- | --- | --- |
| Ramp PVs | N/A | 17.98 |
| Waveforms | 1 | 15.14 |
| Waveforms | 10 | 17.91 | 

## Production Build

| Application Setup | Size of waveform (elements) | Render Time (ms) |
| --- | --- | --- |
| Ramp PVs | N/A | 9.07 |
| Waveforms | 10 | 8.72 |

This suggests that the waveforms are having a limited impact on performance.

Would be a good idea to merge this so that we can test in the future.

Feel free to suggest a better way to factor the WaveformPV in `src/connection/sim.ts` :+1:

aawdls:
For further performance testing we don't want to generate the test data inside the same application. Therefore I am closing this PR but it is a useful reference.

PR #157, opened 2020-03-09
Drop down

TimGuiteDiamond:
Added a drop down/card component :+1: 

Comments welcome :smiley:

### DIFF
--- a/public/json/frontPage.json
+++ b/public/json/frontPage.json
@@ -237,6 +237,53 @@
       ]
     },
     {
+      "type": "dropdown",
+      "position": "relative",
+      "margin": "20px 0 0 0",
+      "title": "Drop Down Section",
+      "children": [
+        {
+          "type": "flexcontainer",
+          "position": "relative",
+          "children": [
+            {
+              "type": "input",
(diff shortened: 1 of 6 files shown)


willrogers:
This is a nice little widget. My main question is: what happens when you put it in an absolute layout?

TimGuiteDiamond:
> This is a nice little widget. My main question is: what happens when you put it in an absolute layout?

Not sure. Will test today.

TimGuiteDiamond:
You can control the width but not the height. See the below screenshot highlighting the parent div, whose size is set from the JSON, but is significantly less tall than the child div inside it:

image

I suppose this would be alright!? With a recommendation that this widget be primarily used within a relative layout or wrapped in a specific display which controls its overflow?

willrogers:
You can set `min-height` on the `<details>` element. Then if you want it to take up a certain amount of space vertically then it will do, although it won't shrink beyond the contents of the summary.

The font needs to be applied to the `<summary>` element rather than the `<details>` element.

But generally I think this is a nice widget and works well.

PR #161, opened 2020-03-17
Widget parsing

willrogers:
Major changes to widget parsing, with the aim of making it more generic.


willrogers:
@TimGuiteDiamond I'll merge this later today unless you have some comments.

TimGuiteDiamond:
Will review this morning

src/types/position.ts
```diff
@@ -22,6 +26,16 @@ export class AbsolutePosition {
...
+    if (
+      invalidSize(x) ||
+      invalidSize(y) ||
+      invalidSize(width) ||
+      invalidSize(height)
+    ) {
+      throw new Error(
+        `Invalid AbsolutePosition (${x},${y},${width},${height})`
+      );
+    }
```
TimGuiteDiamond: This bit seems to work well to pick up errors in the file at runtime 😄

src/ui/widgets/DynamicPage/dynamicPage.tsx
```diff
@@ -36,6 +36,7 @@ export function DynamicPageFetch({
       filetype="json"
       macroMap={map}
       position={new RelativePosition()}
+      defaultProtocol="ca"
```
TimGuiteDiamond: Should we make this a prop?

src/ui/widgets/GroupingContainer/groupingContainer.tsx
```diff
@@ -1,29 +1,33 @@
...
-);
+): JSX.Element => {
+  const style: CSSProperties = {
+    position: "relative",
+    height: "100%",
+    width: "100%",
+    ...props.border?.css()
+  };
+  return <div style={style}>{props.children}</div>;
+};
```
TimGuiteDiamond: This is basically just a box now? Could the same thing be achieved by using `Display` to do this and leave this one with the `fieldset` tag? It was intended to replace the CS-Studio grouping container

willrogers: This is actually a little bit awkward. Although the default look of a grouping container has a group box, this should really be part of the border style rather than the grouping container widget. Other widgets can have a group box style too. However, it is different to other styles of border in that it requires additional elements. 

Many groups in CS-Studio and EDM don't want the border and it messes up the layout if you add it.

As well as all this, the behaviour in bob files is a bit unusual - if you remove the group box style border it changes the x and y coordinates of the contents to compensate.

It needs a bit of thought, but I think this way round is cleaner for the time being.

src/types/position.ts
willrogers: That's the intention. More generally, if we can do validation like this then we have a bit more control.

src/ui/widgets/GroupingContainer/groupingContainer.tsx
TimGuiteDiamond: So in CS-Studio you can have a group box that doesn't have the heading and border?

src/ui/widgets/DynamicPage/dynamicPage.tsx
willrogers: I think so...

src/types/color.ts
```diff
@@ -15,12 +17,20 @@ export class Color {
...
+      // Fall back to the canvas npm package.
+      ctx = createCanvas(100, 100).getContext("2d");
+    }
     if (ctx) {
       ctx.fillStyle = cssColorName;
       return ctx.fillStyle;
     } else {
+      log.warn(
+        `Failed to parse color name ${cssColorName} using canvas element.`
+      );
```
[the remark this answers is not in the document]
willrogers: Yes, I've edited the comment slightly.

src/ui/widgets/GroupingContainer/groupingContainer.tsx
willrogers: Yes, and we often do.

TimGuiteDiamond: I didn't know that 😮

But if you *did* have a group box with a label in CS-Studio, it wouldn't look right here?

willrogers: Right. But this is actually less common for us.

willrogers:
Agreed re. documentation. In principle this structure could work for EDM parsing but I'm not sure what difficulties we might run into with the actual implementation. Probably I need to document this as it is now so that someone else could try to implement that parser.

src/ui/widgets/GroupingContainer/groupingContainer.tsx
TimGuiteDiamond: Okay, in that case, could you map the CS-Studio "grouping container" to the "Display" widget?

I think that they end up being very similar, but the `GroupingContainer` would still be a different, separate thing that we could make use of.

willrogers:
To resume the discussion about grouping containers:

The simplest way for me would be to allow any widget to have any style of border including a group box. Then the grouping container handles grouping things within a display and the display handles things relevant to the root widget of a display. They are conceptually different in my head. Unfortunately it's not immediately straightforward to do this because the group-box-style border requires additional elements, but we can probably find a sensible way of doing that.

We have four similar widgets:

* dynamic page 
* embedded display
* display
* grouping container

What is the dynamic page widget? Your url-info branch is related but I haven't really looked at this part of the code.

TimGuiteDiamond:
In my mind:

- `Display` - the outermost box within which other widgets should be placed. We have often used this at the top of screen files as a root component - because it had different properties to other widgets in our initial attempts to parse JSON into components - but I don't think this is necessary any more. The root component could (I believe, but have not tested) similarly be a `FlexContainer` or `Shape`. *It is not special*.
- `EmbeddedDisplay` - a widget with a defined position which renders a file as children components. *It is special* because it performs the action of fetching the contents of a file from a remote server and renders this file upon reception.
- `DynamicPage` - a widget which may render different pages with different macros at runtime based on the information encoded in the URL. We use this to handle `OPEN_PAGE` and `CLOSE_PAGE` actions, where pages can be opened in a specific `DynamicPage` widget, based on the `routePath` prop. *It is special* because it looks at the URL information at runtime and uses that information to decide which file to render (it uses the `EmbeddedDisplay` widget to perform this rendering).
- `GroupingContainer` - widget for the built in HTML element `fieldset` which looks like a grouping container in CS-Studio (if you have a title). *It is not special*.

NOTE: I have used *special* here to mean *it does something other than directly mapping props into HTML*

Hope this helps clear things up a bit!

willrogers:
I think both the `Display` and the `GroupingContainer` will be special, as in they will have their own functionality. We could have a special widget just to put a group-style border around another widget, but only if we can't find a simpler way of doing it. I wouldn't call it a grouping container though, because that has a meaning already in CS-Studio and Phoebus.

TimGuiteDiamond:
That makes sense.

Can I clarify two things?

- did the above breakdown help to separate the four widgets in your mind?
- what is special about the border for a grouping container in CS-Studio/Phoebus? Perhaps you could send me a simple *.opi file or upload a screenshot? P.S The grouping container seems to be quite different between CS-Studio and Phoebus

What is currently the `GroupingContainer` component could definitely be renamed and as it is a wrapper for a basic HTML element would probably be useful in the future?

willrogers:
* yes
* _what is special about the border for a grouping container in CS-Studio/Phoebus?_ I'm not sure what this question means. In CS-Studio any widget can have a group-box-style border (I think), but in Phoebus it seems only the grouping container allows you special border styles.

TimGuiteDiamond:
Okay, looks like we are mostly on the same page 😄 

If the grouping container border you want is one of these: https://www.w3schools.com/css/css_border.asp then could it not just be applied to the Display widget? Or any other widget for that matter, using the `border` prop? Doe we need a separate widget just for this?

willrogers:
That's how we're doing all the other border styles, it's just the group-box border style including the name that is more difficult.

TimGuiteDiamond:
Does the `GroupingContainer` as on this PR do what you want it to do? I don't see how it includes the title?

willrogers:
That's what I mean by

> Unfortunately it's not immediately straightforward to do this because the group-box-style border requires additional elements, but we can probably find a sensible way of doing that.

and similar comments above.

TimGuiteDiamond:
Ah right!

Well could the `fieldset` and `legend` elements render only when a title is provided?

willrogers:
It could render only when the border style is `GroupBox`:

https://github.com/dls-controls/cs-web-proto/blob/widget-parsing/src/types/border.ts

If we put this logic in the groupingContainer widget then we are a bit further forward, but the logical conclusion is to do this for every widget, leading to a lot of repeated code.

TimGuiteDiamond:
Wow. I really didn't know that this could be added to every single widget 😮

I wonder if the logic for this is ugly in Java too?

Possible solutions:

- we could not support this feature, except for one or two widgets, such as `GroupingContainer`
- if border style is "groupBox" then wrap the internal widget in `fieldset` and `title` (or a `groupingContainer` component). This might work but I think the background colour wouldn't be preserved exactly
- wrap every widget in a special component with this functionality, in a similar manner to `MenuWrapper` and `TooltipWrapper`
- Render opi and bob files grouping containers always with a title (they should have a widget title anyway)

TimGuiteDiamond:
We have spent quite a long time talking about this for what is quite a small detail haha

If you want to keep the `GroupingContainer` to how you have it now, could I ask that you make a new widget with the `fieldset` and `title` elements, as I find them quite nice for layout and design?

I have also added one of the features I mentioned above in #167

Then I think we should put this important step forwards in terms of file parsing into master 👍

willrogers:
> We have spent quite a long time talking about this for what is quite a small detail haha

My observation is that you do not like to leave small details undiscussed. I'll look at finishing this up today.

TimGuiteDiamond:
I think it's worth discussing so that we both understand the issue but I definitely see the value in including this parsing work so we can include it in our other work 👍

PR #162, opened 2020-03-24
Slideshow widget

TimGuiteDiamond:
Added new slideshow widget which allows you to scroll between child widgets, widgets which are not currently shown are not mounted and should therefore not be subscribing to the store.

See the demo at slideshowDemo.json - for example: http://localhost:3000/app/slideshowDemo/%7B%7D/

I haven't quite figured out how to get it to consistently move in the correct direction when I want it to - if you change between moving right and moving left then it looks a bit "wrong" - but I think this is completely possible.


TimGuiteDiamond:
Works well on mobile too!

TimGuiteDiamond:
Rebased to master

willrogers:
This is interesting. I'm happy to merge.

PR #163, opened 2020-03-26
URL Handling

TimGuiteDiamond:
Made some changes to the method used for altering the URL history/path when pages are opened/closed.

I think this provides a slightly more straightforward way of moving data into and out of the URL. This can be seen in the page files where macros as part of the `OPEN_PAGE` action are now normal JSON rather than a string of embedded JSON.

Some of the field names could be changed, but I think the most useful change would be to get the `DynamicPage` component to clean up the URL a bit when it is closed. At the moment, this happens when the close button for the page is pressed, but not when a parent page closes. This leaves unused information hanging around in the URL.

This also provides a good way to allow bob, opi or other file types to be opened directly in a Dynamic page, as the filetype is now included in the information in the URL.


TimGuiteDiamond:
So it turns out it is possible for the `DynamicPage` component to remove itself from the url when unmounting. This is achieved by removing the name frome the url in a function *returned* by a `useEffect` hook with no dependencies.

The issue, is that due to the way `EmbeddedDisplay` is programmed, it has multiple hooks responding to changes on macro, content or filename. Any of these hooks changing triggers a re renders, which causes an embedded `DynamicPage` component to unmount, removing itself from the url and therefore not rendering again 😞 

I don't see an easy way around this, unless we keep some state somewhere which manages the parent-child relationship of the `DynamicPages`.

Nevertheless, I think this is a good change which makes the JSON pages easier to understand.

TimGuiteDiamond:
Have merged later master changes in now. Adding this as I think it makes the page descriptions much easier to understand and is a step in the right direction.

PR #165, opened 2020-04-02
Editor

TimGuiteDiamond:
This is a basic attempt to produce an editor for CS-Web-Proto.

While simple, it does provide an example of what such an editor could look like and does indeed appear to work properly.

It would be nice to properly integrate a JSON viewer/editor, so that it is harder to generate bad JSON. I tried [react-json-view](https://www.npmjs.com/package/react-json-view) briefly but not sure it works exactly for our purposes.

Ultimately it would be great to use the PropTypes specification we have already developed to generate a schema at runtime to help generate the JSON. This could perhaps be investigated separately.

Go to `localhost:3000/app/editor/%7B%7D/` for the demo.

### DIFF
--- a/public/json/editor.json
+++ b/public/json/editor.json
@@ -1,0 +1,4 @@
+{
+  "type": "editor",
+  "position": "relative"
+}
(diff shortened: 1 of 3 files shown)


aawdls:
This will not be merged but it is a useful reference

PR #167, opened 2020-04-09
Improved presentation

TimGuiteDiamond:
My attempts to make our app look better 👍


TimGuiteDiamond:
Rebased against latest changes to master, mainly integration of new urlInfo, and fixed failing unit tests

TimGuiteDiamond:
Notes from @willrogers  - responses in **bold**:

- The toggle theme button should be in the menu
  - Consider trying to implement the alternative theme so it works with the overall theme
  - **Toggle theme button is special at the moment and will need to be wrapped into a widget**
- I can't seem to get back home once I've gone somewhere else
  - **Probably an issue with the link in OPEN_PAGE information**
- The slideshow for the synoptic is interesting but feels experimental. You don't really know when you've opened another screen. If there isn't a screen open the empty slideshow panel doesn't look great. That's all fine though, we're here to some experimenting please carry on (you mentioned tabs for example)
  - **It definitely is experimental at the moment, I have a good plan of haw how to handle this but it will require adding some further functionality; I am thinking along the lines of the dynamicPage component which can extract information from the url, but with additional state so it can open/close tabs as requested**
- I might expect the drawer to close when I select one of the items
  - **I think this can be accommodated without too much hassle but would need to look at the docs for the component**
- I'm not sure the slider widget is working properly at the moment. This certainly might have been one of my changes
  - **This has been broken for sometime but should be fixed, yes**
- Also, can you put a note on the Wiki about how to serve the development app on the local network. I guess you're doing this so you can use your phone for testing.
  - **Yes, I will and it does look decent on my phone** 👍

src/app.tsx
```diff
@@ -57,25 +57,43 @@ const App: React.FC = (): JSX.Element => {
...
+              )
+            }
+          />
+          <Provider store={store}>
+            <div className="App">
+              <button type="button" onClick={toggle}>
+                Toggle Theme
+              </button>
+              <Profiler id="Dynamic Page Profiler" onRender={onRenderCallback}>
+                {/* <DynamicPageWidget
```
willrogers, line 83: There's some commented code left behind here.

PR #169, opened 2020-05-05
Complete macro refactor.

willrogers:
The main changes are:

* only container widgets now have a macro prop
* it is possible to change a display's macros from inside the display
* the MenuMux widget allows changing the display's macros

This uses React Context rather than props to implement the macros.


TimGuiteDiamond:
Hey Will, I haven't really seen a `MenuMux` in use before I don't think - it looks like it allows you to dynamically set macros? Could you put an example of this in the `mobileLayout.json`?

willrogers:
Your innocent little question exposed a bug, which I have now fixed.

TimGuiteDiamond:
Glad to hear that! It seems to work well 😃
I would still like to take a closer look at the code (have been working on benchmarking code today) but if you are happy with this then we can merge and I will look later.

willrogers:
See also https://github.com/dls-controls/cs-web-proto/wiki/Macros

Issue #168, opened 2020-05-05
Tabbed Container


TimGuiteDiamond:
In an effort to improve the style and usability of our web application, supporting existing workflows based around the design of beamline screens in EDM and CS-Studio, I am suggesting here a specification for a tabbed container widget.

# Wireframe

See below wireframes with comments on how such a widget might work and how it could incorporate the beamline workflow.

![image](https://user-images.githubusercontent.com/50404435/81073712-f48db400-8edf-11ea-8230-d801a2ebc847.png)

![image](https://user-images.githubusercontent.com/50404435/81073756-02433980-8ee0-11ea-83f2-d1cc3526fdfd.png)

![image](https://user-images.githubusercontent.com/50404435/81073781-0cfdce80-8ee0-11ea-9cb8-c8b4931d4a8d.png)

# How does it work in EDM?

It is unlikely we will be able to, or even desire to replicate the EDM workflow.
EDM works by opening up a new application window for every component screen and detail screen, although it does put some effort into ensuring that a screen can exists as a singleton - i.e. pressing a button which opens a screen repeatedly will only create a singleton copy of the screen, and will bring it to the forefront of the display.

The `Related Display` widget which opens new screens allows the filename and macros to be set, along with other properties:

![image](https://user-images.githubusercontent.com/50404435/81074908-906bef80-8ee1-11ea-9e63-edf623e5add9.png)

Key macros are used to populate the new window, including the window title:

![image](https://user-images.githubusercontent.com/50404435/81075018-b5f8f900-8ee1-11ea-81cf-a7ef04dc4baf.png)

As the windows are operating separately, they can be safely closed in most instance without disrupting other screens.
This method also avoids the need to describe *where* a screen should open, as a completely new window will be created.

# How does it work in CS-Studio/Phoebus?

CS-Studio is built on the Eclipse framework, so comes with some built in methods for handling tabs, and targetting different areas of the screen.
The DLS-specific setup has split the screen up into sections, allocated to the synoptic, component and details screens separately.
However, this is by convention only.
We can see that when specifying the open OPI action, a limited number of possiblities are provided:

![image](https://user-images.githubusercontent.com/50404435/81075603-7aaafa00-8ee2-11ea-9a69-caf1d25ed0aa.png)

Phoebus also allows a limited number of options:

![image](https://user-images.githubusercontent.com/50404435/81076207-4be15380-8ee3-11ea-941f-f7406304b790.png)

[...]

PR #171, opened 2020-05-20
Tabs

TimGuiteDiamond:
PR making progress on #168


src/ui/widgets/Tabs/tabContainer.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  console.log(props.tabs);
+
+  try {
+    // TODO: Find out if this repeated calculation can be done in the useMemo hook for measurable performance gains
+    const children = Object.values(props.tabs).map((child, index) =>
+      widgetDescriptionToComponent(
+        parseJson(JSON.stringify(child), "pva"),
+        index
+      )
```
TimGuiteDiamond, line 42: @willrogers I don't feel great about stringifying this object just to parse it again, but I think it is acceptable at the moment. If we start finding more places where this pops up then perhaps we should look for another way to do it.

This method keeps the JSON parsing clean (other solutions I thought of involved changing the JSON parser to check for the special case of tabs).

Let me know if you think there is a better way to do this.

TimGuiteDiamond:
# Handling Recursion

There is a potential issue with tabs and embedded screens in general - what happens when a tab requests that the file containing it be shown inside it's body?

We can see that this will quickly lead to an infinite loop, which is usually considered a bad thing.

So, what is the best way to handle this? Let's take a look at the existing display managers:

# Phoebus

Initially, Phoebus handled this relatively well and seemed to be fine, however it clearly does not have a maximum limit of any kind and so after scrolling down a bit to find the innermost copy of the file, it failed.

image

No warning message was displayed in Phoebus itself, but error messages can be seen on the console.

image

# CS-Studio

CS-Studio renders the first level of recursion, but then notes that there is a loop and refuses to continue.
The application remains functional and there is a useful warning message within the application console.

image

Warning message:

image

# EDM

EDM takes a bit of coaxing to do this but the Embedded Window widget does the job.
Similarly to Phoebus, it does not fail obviously, although it gradually becomes less and less responsive.

image

When trying to exit the program, an error is shown. The program must be killed at the terminal.

image

# CS-Web-Proto

So, what should we do? I think CS-Studio handles this the best - perhaps this logic will find it's way into Phoebus soon enough as well.
We need to think about how this would be handled.

TimGuiteDiamond:
I've now made some progress on defining the actions for handling opening and closing of tabs. The rest of the implentation should be fairly straightforward. Then it will be a case of taking the basic structure of the `navigationTabs` component and providing it with a connection to the URL and close buttons which perform the `CLOSE_TAB` action.

Should also clean up actions by renaming `page` in OPEN_PAGE to `location`

TimGuiteDiamond:
Rebased of latest changes to `master` which should include tartiflette

TimGuiteDiamond:
@willrogers please take a look at this and `tabbed-synoptic` branch - I think both could be merged onto master.

I will separately begin work on aligning the different dynamic page components and actions as we have discussed.

TimGuiteDiamond:
Merged onto #176 which continues this work

PR #173, opened 2020-06-18
Tartiflette

willrogers:
@skyfrench a first version that works with the tests passing.

A lot of this was written fairly quickly, so a review would be handy. I don't think there are enough comments, as well. I suggest that you start from this branch, so you can find the tricky bits before we merge this branch.


skyfrench:
> @skyfrench it's a huge diff, so I suggest you look at `coniql.ts` rather than trying to review the whole thing.

Thanks @willrogers - I'll take a look.

PR #176, opened 2020-07-02
Dynamic content

TimGuiteDiamond:
Builds on the changes made in #171 and aims to align the different ways of opening pages dynamically.

## Main aims
- Give all open and close actions the same shape for both pages and tabs
- Provide simpler names for the information in the above actions
- Use all of the dynamic information as a key to the screen for dynamic tabs - this means the same screen can be opened with different macros in different tabs, possibly with the same tab title

## Challenges for another day
- Moving information from URL to context or Redux
- One pair of open/close actions for both pages and tabs - this will require runtime knowledge of whether a container is a page or a tab and might be achievable as part of, or after, the above point


TimGuiteDiamond:
Proposed type to be used for dynamic content:

```typescript
interface dynamicContent {
  name: string; // Name associated with the content
  location: string; // Location of component to target
  description?: string; // Optional description of action
  file: {
    // All information required for an embedded display
    path: string; // Name or file of path (without suffix ?)
    type: "json" | "opi" | "bob"; // File type - which parser and suffix
    macros: { [key: string]: string }; // Macros
    defaultProtocol: string; // Default PV prefix for parser
  };
}
```

### Notes

- `location` is called `Mode` in CS-Studio, but ours is more general
- the suffix for the file is almost always the same as the `type` but requiring the suffix to be excluded so we can add it is a likely source of errors
  - someone passes `path` of `myfile.bo` and `type` of `opi` -> assumptions made by us could cause problems
  - easiest and simplest would be to require both full `path` and `type` - although this could still cause errors if users do not specify the correct parser?
- `name` is currently required - in the future this could perhaps be specified in the file as in CS-Studio

TimGuiteDiamond:
## Goals one and two have been completed

- Give all open and close actions the same shape for both pages and tabs
- Provide simpler names for the information in the above actions

## Goal three requires further work

- Use all of the dynamic information as a key to the screen for dynamic tabs - this means the same screen can be opened with different macros in different tabs, possibly with the same tab title

Probably we need a specialised function to compare dynamic information against each other to avoid both the shallow comparison and stringify problem. I see the "stringify problem" being that the same specification could be written with the keys in a different order, leading to another new tab where there shouldn't be one.

If we are going to use all this information as a key, then we probably don't need a key/value setup, as the value would be redundant. However, we would still like to be able to address tabs or pages quickly, to ascertain whether it has already been described or not. A `set` would be useful for this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

PR #183, opened 2020-09-29
Documentation

GeorgeSheppard:
Mostly documentation additions, some minor code changes to simplify code in app, and themeContext. Addition of a few more tests all around.

### DIFF
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -13,7 +13,7 @@ import {
   FileContextType,
   LocationCache,
   FileDescription,
-  fileDescEqual,
+  fileDescEqual
 } from "./fileContext";
 import { Header } from "./ui/components/Header/header";
 import { Footer } from "./ui/components/Footer/footer";
@@ -35,9 +35,9 @@ const App: React.FC = (): JSX.Element => {
         path: "home.json",
         type: "json",
         macros: {},
-        defaultProtocol: "pva",
-      },
-    ],
+        defaultProtocol: "pva"
+      }
+    ]
   });
 
   // Manages locations, and functions for adding and removing location
@@ -55,7 +55,7 @@ const App: React.FC = (): JSX.Element => {
         delete locationsCopy[location];
       }
       setLocations(locationsCopy);
-    },
+    }
   };
 
   return (
@@ -75,7 +75,7 @@ const App: React.FC = (): JSX.Element => {
                   path: "app.json",
                   type: "json",
                   defaultProtocol: "pva",
-                  macros: {},
+                  macros: {}
                 }}
               />
             </Profiler>
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -13,6 +13,7 @@ import {
   FileContextType,
   LocationCache,
   FileDescription,
+  fileDescEqual,
 } from "./fileContext";
 import { Header } from "./ui/components/Header/header";
 import { Footer } from "./ui/components/Footer/footer";
@@ -21,7 +22,9 @@ const baseUrl = process.env.REACT_APP_BASE_URL ?? "http://localhost:3000";
 
 log.setLevel((process.env.REACT_APP_LOG_LEVEL as LogLevelDesc) ?? "info");
 
+// React.FC: React functional component
 const App: React.FC = (): JSX.Element => {
+  // Set dark or light mode using ThemeContext
   const { dark, applyTheme } = React.useContext(ThemeContext);
   applyTheme(dark ? darkTheme : lightTheme);
 
@@ -37,22 +40,28 @@ const App: React.FC = (): JSX.Element => {
     ],
   });
 
+  // Manages locations, and functions for adding and removing location
   const fileContext: FileContextType = {
-    locations: locations,
+    locations,
     addFile: (location: string, desc: FileDescription, name: string) => {
       const locationsCopy = { ...locations };
       locationsCopy[location] = [name, desc];
       setLocations(locationsCopy);
     },
     removeFile: (location: string, desc: FileDescription) => {
-      // TODO: match the description.
       const locationsCopy = { ...locations };
-      delete locationsCopy[location];
+      // description stored in second element of array
+      if (fileDescEqual(desc, locationsCopy[location][1])) {
+        delete locationsCopy[location];
+      }
       setLocations(locationsCopy);
     },
   };
 
   return (
+    // Each instance of context provider allows child components to access
+    // the properties on the object placed in value
+    // Profiler sends render information whenever child components rerender
     <FileContext.Provider value={fileContext}>
       <BaseUrlContext.Provider value={baseUrl}>
         <Provider store={store}>
@@ -60,6 +69,7 @@ const App: React.FC = (): JSX.Element => {
             <Header />
             <Profiler id="Dynamic Page Profiler" onRender={onRenderCallback}>
               <EmbeddedDisplay
+                // RelativePosition returns CSS properties
                 position={new RelativePosition()}
                 file={{
                   path: "app.json",
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -22,7 +22,6 @@ const baseUrl = process.env.REACT_APP_BASE_URL ?? "http://localhost:3000";
 
 log.setLevel((process.env.REACT_APP_LOG_LEVEL as LogLevelDesc) ?? "info");
 
-// React.FC: React functional component
 const App: React.FC = (): JSX.Element => {
   // Set dark or light mode using ThemeContext
   const { dark, applyTheme } = React.useContext(ThemeContext);
@@ -50,8 +49,13 @@ const App: React.FC = (): JSX.Element => {
     },
     removeFile: (location: string, desc: FileDescription) => {
       const locationsCopy = { ...locations };
-      // description stored in second element of array
-      if (fileDescEqual(desc, locationsCopy[location][1])) {
+
+      // description stored in second element of array but may not exist
+      if (locationsCopy[location][1] !== undefined) {
+        if (fileDescEqual(desc, locationsCopy[location][1])) {
+          delete locationsCopy[location];
+        }
+      } else {
         delete locationsCopy[location];
       }
       setLocations(locationsCopy);
--- a/src/app.tsx
+++ b/src/app.tsx
@@ -12,7 +12,7 @@ import {
   FileContext,
   FileContextType,
   LocationCache,
-  FileDescription
+  FileDescription,
 } from "./fileContext";
 import { Header } from "./ui/components/Header/header";
 import { Footer } from "./ui/components/Footer/footer";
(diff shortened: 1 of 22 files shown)


willrogers:
It looks like you're using jsdoc conventions for documenting functions. This seems like a reasonable convention (it would take a lot of work to make all exported values conform, but we don't have to do that).

If so we should mention it in the development environment wiki page.

src/types/font.test.ts
```diff
@@ -21,4 +21,9 @@ describe("Font", (): void => {
       fontStyle: "italic"
     });
   });
+
+  it("size is left out of CSSProperties when not input", (): void => {
```
willrogers, line 25: The comment says size but the test says style...?

src/types/font.ts
```diff
@@ -20,12 +23,18 @@ export class Font {
     typeface?: string,
     name?: string
   ) {
+    // All parameters are optional but typeface and style have a default
+    // ?? returns the right expression if the left is null or undefined
```
willrogers, line 27: This is an example of documenting syntax.

src/types/macros.ts
```diff
@@ -5,22 +5,25 @@ export interface MacroMap {
...
+ * are the same
+ * @param first first MacroMap object
+ * @param second second MacroMap object
+ * @returns boolean, true if they match else false
+ */
 export function macrosEqual(first: MacroMap, second: MacroMap): boolean {
-  // TODO why are the DIDs not matching?
   const { DID: firstDid, ...firstOthers } = first;
-  const { DID: secondDid, ...secondOthers } = first;
+  const { DID: secondDid, ...secondOthers } = second;
```
willrogers, line 17: Looks like you've caught another bug here. Is this one tested?

src/types/macros.ts
```diff
@@ -5,22 +5,25 @@ export interface MacroMap {
...
+
   for (const key in firstOthers) {
-    if (first.hasOwnProperty(key)) {
-      if (first[key] !== second[key]) {
-        return false;
-      }
+    if (first[key] !== second[key]) {
```
willrogers, line 20: Again, there's a reason why these lines of code were there, though arguable.

src/types/macros.test.tsx
```diff
@@ -20,3 +21,40 @@ it("doesn't substitute if no macro matches", (): void => {
   const substituted = resolveMacros("X${F}", macroMap);
   expect(substituted).toEqual("X${F}");
 });
+
+it("varying brackets are still substituted", (): void => {
```
willrogers, line 25: I really like this test. Exposes questionable behaviour.

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -60,30 +60,35 @@ export const EmbeddedDisplay = (
...
-    return (): void => {
-      mounted = false;
-    };
-  });
+      // Clean up function
+      return (): void => {
+        mounted = false;
+      };
+    }
+    // TODO 28/09/20: Course I did said should always declare dependencies as second argument
```
willrogers, line 88: Good question, we should think about this.

willrogers (commented):
This is great stuff, you've evidently got a good understanding already. There is a bit of a mix in here though, not just documentation:

* comments: mostly great
* addtional tests: this is a great way to test your understanding of the code, excellent
* syntax tidying: this is useful and can be added in a documentation branch reasonably enough
* bug fixes: here I think you have fixed a couple, others you may have changed behaviour a bit. Some could have tests, each should at least have its own commit

You'll find my own changes in this project are similar to yours, it is a prototype and if something breaks then it's probably a sign that the code isn't clear and should be improved anyway. I'll go through and make a few comments.

GeorgeSheppard:
> It looks like you're using jsdoc conventions for documenting functions. This seems like a reasonable convention (it would take a lot of work to make all exported values conform, but we don't have to do that).
> 
> If so we should mention it in the development environment wiki page.



> This is great stuff, you've evidently got a good understanding already. There is a bit of a mix in here though, not just documentation:
> 
>     * comments: mostly great
> 
>     * addtional tests: this is a great way to test your understanding of the code, excellent
> 
>     * syntax tidying: this is useful and can be added in a documentation branch reasonably enough
> 
>     * bug fixes: here I think you have fixed a couple, others you may have changed behaviour a bit. Some could have tests, each should at least have its own commit
> 
> 
> You'll find my own changes in this project are similar to yours, it is a prototype and if something breaks then it's probably a sign that the code isn't clear and should be improved anyway. I'll go through and make a few comments.

That makes sense to split it up, should I be putting bug fixes on an entirely different branch to make it clear? If so should I just leave a TODO on the documentation branch and come back to it once it's been merged in?

src/app.tsx
```diff
@@ -21,15 +22,10 @@ const baseUrl = process.env.REACT_APP_BASE_URL ?? "http://localhost:3000";
...
 log.setLevel((process.env.REACT_APP_LOG_LEVEL as LogLevelDesc) ?? "info");
 
-function applyTheme(theme: any): void {
-  Object.keys(theme).forEach(function(key): void {
-    const value = theme[key];
-    document.documentElement.style.setProperty(key, value);
-  });
-}
-
+// React.FC: React functional component
```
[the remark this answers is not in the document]
GeorgeSheppard: I've never thought about it like that before as it's something I do quite often, but documenting intent makes more sense. Fixed.

willrogers:
I don't think it needs to be a different branch, more that this branch happened to be called documentation.

Different commits are a good idea though as when you look at some code and don't understand it, it's useful to look at the particular commit to figure out why the change was made.

src/app.tsx
```diff
@@ -43,29 +39,37 @@ const App: React.FC = (): JSX.Element => {
...
+    locations,
     addFile: (location: string, desc: FileDescription, name: string) => {
       const locationsCopy = { ...locations };
       locationsCopy[location] = [name, desc];
       setLocations(locationsCopy);
     },
     removeFile: (location: string, desc: FileDescription) => {
-      // TODO: match the description.
       const locationsCopy = { ...locations };
-      delete locationsCopy[location];
```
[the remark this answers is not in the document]
GeorgeSheppard: Sure, I've added an if that checks for an undefined element that falls back on the previous logic, should mean the behavior is as before in that case but fixing when we encounter a problem works with me.

src/index.tsx
```diff
@@ -5,6 +5,7 @@ import App from "./app";
 import { ThemeProvider } from "./themeContext";
 
 ReactDOM.render(
+  // Website wrapped in ThemeProvider so dark and light mode is global
```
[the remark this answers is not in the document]
GeorgeSheppard: Yes, I will change it to available anywhere as it's clearer.

src/themeContext.tsx
```diff
@@ -1,50 +1,66 @@
...
 
+/**
+ * Switches between the dark and light theme and stores user
+ * preference, toggle accessed using
+ * const { toggle } = useContext(ThemeContext)
+ */
 export const ThemeProvider: React.FC<{}> = ({ children }: any): JSX.Element => {
-  const [dark, setIsDark] = useState(false);
-
-  useEffect((): void => {
```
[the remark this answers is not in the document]
GeorgeSheppard: I will make sure to do that next time, effectively after the first render useEffect was a redundant function as the toggle button was storing and setting dark to it's new value, at which point useEffect was running and reading and setting dark to the value it was already set to. UseEffect I just replaced with one line (47).

src/types/dtypes.ts
```diff
@@ -160,11 +160,16 @@ export class DType {
...
+   * Tries to convert a DType into a string by first converting to
+   * an intermediary value (first tried is string, then double, then array)
+   * @param value
+   */
   public static coerceString(value?: DType): string {
     if (value) {
       const stringValue = value.getStringValue();
       const doubleValue = value.getDoubleValue();
-      const arrayValue = value.getDoubleValue();
+      const arrayValue = value.getArrayValue();
```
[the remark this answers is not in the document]
GeorgeSheppard: Never even thought about adding a test, but that makes sense, have added it now.

src/types/dtypes.ts
```diff
@@ -204,6 +214,11 @@ export class DType {
     return this.value.arrayValue;
   }
 
+  /**
+   * Attempts to extract a arrayValue from a DType through an
+   * intermediary conversion (first tried is array, then double)
+   * @param value
```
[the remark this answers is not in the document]
GeorgeSheppard: I agree, but it looked like it was boilerplate code that wasn't going to need changing so I thought it would be worth it even just so I can see the function description when I hover over the function in other files.

src/types/font.test.ts
GeorgeSheppard: Good catch

src/types/macros.test.tsx
GeorgeSheppard: Thank you! Really messes with my bracket colourer in vscode though

src/types/macros.ts
GeorgeSheppard: It's tested now.

src/types/macros.ts
GeorgeSheppard: My thinking was that it is iterating over the keys in firstOthers, and as firstOthers is a subset of first all keys in firstOthers are guaranteed to be in first (as long as the above bug where first is assigned to both lines doesn't happen again [which is now tested]). If there is another reason I am missing please let me know though.

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
GeorgeSheppard: Looking at it again, I would think it would be 'file' as the second dependency, I don't know the big picture though so up to you if you think that will work.

src/types/macros.ts
willrogers: Oh I see. These were actually supposed to be `firstOthers.hasOwnProperty(key)`. It's all to do with a subtlety in how objects are compared in JavaScript. See https://stackoverflow.com/a/5859028 for an example.

However, we may well never hit the corner cases that is trying to allow for, so I'm not really inclined to change this back.

willrogers:
OK, generally looks good now. As I say this is a prototype so let's merge and fix the problems we find.

One final comment on commits: the ideal situation is that I see a line of code that I don't understand, I dig out which commit it came from and I find a commit that made one coherent change and a commit message that says what the change is and why. The last commit you made isn't that!

Otherwise, good work, thanks!

GeorgeSheppard:
> OK, generally looks good now. As I say this is a prototype so let's merge and fix the problems we find.
> 
> One final comment on commits: the ideal situation is that I see a line of code that I don't understand, I dig out which commit it came from and I find a commit that made one coherent change and a commit message that says what the change is and why. The last commit you made isn't that!
> 
> Otherwise, good work, thanks!

I will take that on board, I'm not used to using git properly like this so please keep pointing out when I don't use it properly. Thank you.

PR #184, opened 2020-09-30
groupingContainer and groupBox were both exporting functions called G…

GeorgeSheppard:
groupingContainer and groupBox were both exporting functions called GroupingContainer, they were both registering with the same name as well. The names both widgets register with and the name of the exported functions have been changed.

### DIFF
--- a/src/ui/widgets/GroupBox/groupBox.tsx
+++ b/src/ui/widgets/GroupBox/groupBox.tsx
@@ -31,8 +31,8 @@ const GroupBoxProps = {
   children: ChildrenPropOpt
 };
 
-export const GroupingContainer = (
+export const GroupBox = (
   props: InferWidgetProps<typeof GroupBoxProps>
 ): JSX.Element => <Widget baseWidget={GroupBoxComponent} {...props} />;
 
-registerWidget(GroupingContainer, GroupBoxProps, "grouping");
+registerWidget(GroupBox, GroupBoxProps, "groupingBox");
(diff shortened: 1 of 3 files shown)


src/ui/widgets/GroupBox/groupBox.tsx
```diff
@@ -31,8 +31,8 @@ const GroupBoxProps = {
   children: ChildrenPropOpt
 };
 
-export const GroupingContainer = (
+export const GroupBox = (
   props: InferWidgetProps<typeof GroupBoxProps>
 ): JSX.Element => <Widget baseWidget={GroupBoxComponent} {...props} />;
 
-registerWidget(GroupingContainer, GroupBoxProps, "grouping");
+registerWidget(GroupBox, GroupBoxProps, "groupingBox");
```
willrogers, line 38: probably "groupBox" for consistency.

willrogers (commented):
Well spotted.

Check `src/ui/widgets/index.ts`, I think we may need a change there as well.

Also, where's the page describing how to write a widget, let's make sure it's up to date.

willrogers:
https://github.com/dls-controls/cs-web-proto/wiki/How-to-make-a-Widget

Have a read, I suspect it is a little out of date. I think you will be making a few widgets.

GeorgeSheppard:
> https://github.com/dls-controls/cs-web-proto/wiki/How-to-make-a-Widget
> 
> Have a read, I suspect it is a little out of date. I think you will be making a few widgets.

Will do, will finish the documentation for the current widgets first thing tomorrow morning, then start on that.

PR #186, opened 2020-10-07
Led widget

GeorgeSheppard:
Creating an LED widget, including tests, and reformatting Readback alarm code into a common function for any widget that needs alarm functionality. Added an extra page to test individual widgets on.


willrogers:
Thanks, looks much better now that the commits are split up separately. The widget looks nice - now what we need to do is compare how it functions with the LED widget in CS-Studio. I guess you're going in this direction from your comments on Slack.

I wouldn't have put the css refactor on this branch, because now we've got two lots of things to discuss here.

GeorgeSheppard:
> Thanks, looks much better now that the commits are split up separately. The widget looks nice - now what we need to do is compare how it functions with the LED widget in CS-Studio. I guess you're going in this direction from your comments on Slack.
> 
> I wouldn't have put the css refactor on this branch, because now we've got two lots of things to discuss here.

No worries, I agree. I didn't create it myself just found it online and modified it slightly. Yes I'm going that direction but need help finding the files. In terms of the css refactor, I've not made a specific choice on this branch, and have left the getClass function such that you can specify the default value of the class instead of it attempting to access a css property with the name Default. An example of this is in readback.tsx line 76 where it specifies the Readback styling as the default whereas in led.tsx line 26 no styling is specified so it extracts the Default styling.

willrogers:
OK, well we'll deal with the CSS later.

Have you managed to launch CS-Studio yourself? If so, you could create a new opi file, add an LED widget and see what options it gives you. This will be good practice anyway, as CS-Studio shows a lot of the behaviour that we want to recreate.

I don't know if we use LEDs more for _showing the error state of a PV_ or _indicating the value of a PV_.

GeorgeSheppard:
I will give that a go, that is very useful to know, I will add another commit if it needs changing so feel free to leave this pull open till then.

GeorgeSheppard:
@willrogers this should be done now. I believe the LEDs are used more for user defined rules, so I've added this in.

GeorgeSheppard:
Actually hold off on that, I am working on a simpler user definition.

willrogers:
We need to use a similar structure to the existing widgets. So I think they set the LED colour using a rule, but don't have special props like you have `rule` and `useRule`.

This is why I'm very keen to load existing screens, to make sure what we're doing is compatible.

GeorgeSheppard:
That's what I was just thinking, I can get rid of useRule by checking if any rules are present on the props, and rule isn't a user parameter it is just the value that gets passed back when the rule is resolved so I can't imagine it being a problem unless I am missing something. Readback uses backgroundColor to pass it back instead, which I suppose I could use to set the color of the LED so that they can use any color type actually. So I will rework it to be more similar to Readback.

PR #187, opened 2020-10-07
Use effect

willrogers:
@GeorgeSheppard your comment required a few changes and subsequently I did a lot of test refactoring, which is coming in a subsequent PR.

### DIFF
--- a/src/setupTests.tsx
+++ b/src/setupTests.tsx
@@ -1,6 +1,5 @@
 // React testing library extensions to expect().
 import "@testing-library/jest-dom/extend-expect";
-// Set up Enzyme.
 import { configure } from "enzyme";
 import log from "loglevel";
 import Adapter from "enzyme-adapter-react-16";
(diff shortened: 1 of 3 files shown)


GeorgeSheppard (approved):
spyOn looks interesting, do you think that could be used to remove some of the redundant logging when running tests (in sim.test.ts for example)?

willrogers:
I'm not sure what you mean by this, but I suggest that if logging is redundant then the logging itself should be reviewed.

GeorgeSheppard:
I mean that, when running the tests there is sometimes logging that is a byproduct of running a function, and this just clogs up the terminal and distracts from warnings. For instance when running 'npm test' the functions in sim.ts log when unsubscribing from a PV, I was thinking that by using spyOn on console.log then the logs could be stopped when running tests and that would make it clearer when looking through the test results. I'm not sure I'd class the logging as redundant, just not useful when running tests.

PR #191, opened 2020-10-09
Adding led

GeorgeSheppard:


GeorgeSheppard:
@willrogers I have updated the alarm sensitivity for the default LED which just monitors the alarm level on a PV, but for user rules the functionality is slightly different. The user defines rules and colours to return in the relevant .json file, much like how this can be done for the "label" widget (mobileDemo.json line 28). Have a look on the widgetTestPage if you want to see both types side by side.

willrogers:
I'm just trying to figure out how the LED widget in CS-Studio does this.

willrogers:
Answer: it's complicated. Could you make this simpler though: one configured colour, then:

* if alarm sensitive, change it to red or yellow or white depending on the alarm quality
* alternatively set a rule to change the colour property

? 

This raises the question of which takes precedence if both are set. The rule, I think, which is probably appropriate.

GeorgeSheppard:
I think what you mean was basically what it did have before, but it didn't have the alarmSensitive property so I have added that. 
Configured colour is green
User rule takes precedence over alarm value, even if alarmSensitive is set to true
If alarmSensitive and no user rule changes to yellow, red, or white (have taken the colours from your previous PR)

willrogers:
OK, let's see how we get on with it.

The other thing I don't understand is the `scale` prop. Does it scale the size of the LED? Wouldn't you use the widget size to control the size of the LED?

GeorgeSheppard:
Sounds good, and yes it does, there's a simple reason and it's I didn't know a widget size prop existed. I will have a look into it.

willrogers:
Widgets all have a size, either absolute or relative. I'd expect an LED to fill its container, similar to what I think you were playing with for the SVGs.

GeorgeSheppard:
Yes looks like the .opi has 'width' tied to each of the LED's, so can use that.

PR #200, opened 2020-10-20
Making sizes in web more similar to sizes in CS-Studio, by subtractin…

GeorgeSheppard:
Minor update, to make the LED in the web look more similar to the size in CS-Studio, a fixed value of 5 pixels is subtracted from the value received from the .opi file. Tests have then been updated.

### DIFF
--- a/src/ui/widgets/LED/led.module.css
+++ b/src/ui/widgets/LED/led.module.css
@@ -1,6 +1,6 @@
 .Led {
-  width: 16px;
-  height: 16px;
+  width: 11px;
+  height: 11px;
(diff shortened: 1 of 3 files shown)


GeorgeSheppard:
I should add that I don't like having to subtract 5 pixels like this, but the slightly more complicated design of the LED in CS-Studio means the sizes are out of sync. In CS-Studio the LED has a small border that transitions between the LED and the background (and causes this size discrepancy), which is on my TODO list to implement so I don't have to subtract these 5 pixels like this.

PR #201, opened 2020-10-20
Symbol widget

GeorgeSheppard:


willrogers:
I think ideally what we would have here is that the image component is capable of displaying an SVG and then the symbol component uses the image component.

How did you manage to get the SVGs the right size in the end?

willrogers:
I think ideally what we would have here is that the image component is capable of displaying an SVG and then the symbol component uses the image component.

How did you manage to get the SVGs the right size in the end?

GeorgeSheppard:
I agree just wanted to run it past you first, it was actually quite simple and just needed to set the height and width properties in the style prop. So I could just use the 'fill' prop on the image component to determine whether to use the 100% fill or to use the real size of the image?

By the way the comment I deleted of yours is just a duplicate.

willrogers:
Again, I suggest you look at the properties on the widgets in CS-Studio to see how they manage it.

GeorgeSheppard:
So looking at CS-Studio the conclusion I have is this: CS-Studio doesn't have a method for handling it, because there is no such thing as an image widget in it (at least not on the synoptic). So my personal opinion is that the SVG image component and the image component can be combined into one, with the default behaviour being the current SVG image behaviour. The only time the current image widget is being used is in custom json files and never from CS-Studio, so I think it makes sense in terms of compatibility to have the current SVG widget I made as the default, and then when defining a image widget in one of the jsons without a height and width parameter it can default back to the current fill ehaviour.

willrogers:
There is an image widget in CS-Studio. In the editor you should be able to add new widgets from the right-hand side, there are lots of them.

GeorgeSheppard:
Ahh I didn't see there is a palette with all the widgets on that makes life easier. Looking at it, it has the same behaviour that the SVG image component has, so I still think that using the SVG behaviour is the best option. I think the image component (as it is used currently) with the autofill behaviour isn't the behaviour used by the image widget or the symbol widget.

willrogers:
I just checked the image widget in CS-Studio and its behaviour is fairly simple, although again simplified a bit by the fact that opibuilder uses absolute positioning:

* displays an image (png, svg and gif are supported; I didn't test jpg)
* by default, it will resize the widget to the size of the image itself
* you can choose to resize the widget
* you can choose to stretch the image to the resized widget
* you can rotate the image (although only to 90, 180 or 270 degrees)
* you can flip it or crop it

This makes sense to me and I think our image widget should have similar functionality, although we can add particular features as we need them.

The symbol widget has the particular feature that you can display a different image depending on the value of a PV. It has an awkward way of figuring out the name of the different images though. I think at the moment this symbol widget just shows one image. It also allows you to put a label on the symbol, although as far as I can tell that must come from the PV.

willrogers:
I would have thought that once we have an image widget with those capabilities, we would be able to use a component from that in the symbol widget to do the image rendering and save ourselves a bit of repetition.

GeorgeSheppard:
@willrogers I have added a good amount of the features you outlined above. Flips, arbitrary rotations, resizing of the widget, and stretching of the image are done. The method for handling the stretchToFit props, however, is very hacky.

src/ui/widgets/Symbol/symbol.tsx
```diff
@@ -0,0 +1,90 @@
...
+        }}
+      />
+      {showLabel && (
+        <LabelComponent
+          {...{ visible, backgroundColor: background, ...props }}
+          text={props.value?.getStringValue() ?? ""}
+          // TODO: This is pretty hacky, on CS-Studio when using stretchToFit
+          // the text stays still while the image expands underneath, to compensate
+          // for expansion of the image the text is shifted a FIXED amount
+          transform={`translate(0px, ${props.stretchToFit ? -20 : 0}px)`}
```
GeorgeSheppard, line 73: @willrogers here is the hacky code, I know it is bad but to do it properly requires the size of the original image to determine the effect stretchToFit will have on the text location. Do we care enough about this specific feature to implement it properly? If there is an easy method I am missing then let me know, but otherwise from the amount stretchToFit is used in CS-Studio I would assume not.

src/ui/widgets/Image/image.tsx
```diff
@@ -43,8 +51,11 @@ export const ImageComponent = (
         src={file}
         alt={props.alt || undefined}
         style={{
-          height: imageSize,
-          width: imageSize
+          height: props.height ?? imageSize,
+          width: props.width ?? imageSize,
```
willrogers, line 55: Here we have a problem. If the widget width is e.g. `50%`, it will have a container div of that size but this 50% will be 50% of that container so will end up as 25%. Try setting 50% on your test widget.

What were you aiming for with these lines?

willrogers: What do you think the `fill` prop was aiming for? Stretch to fit? We should change it to align with the CS-Studio props.

willrogers: I mean change the name.

GeorgeSheppard: You are right it does make it 25%, but I don't understand why that is happening at all. My goal was just to use the defined height if it is available otherwise default back to the old behaviour. I haven't changed the old behaviour so why is that happening?

I can change the name for the fill prop I assumed it was aiming to just display the image with it's normal size, that makes my other comment redundant which is good.

GeorgeSheppard: I'm not sure if I have misunderstood you, but the fill prop does not have the same result as the stretchToFit does in CS-Studio.

willrogers: Try `height: "300px", width: "300px"` in the diamond logo on the widget test page.

willrogers: This is not a simple corner of the code, because of the way CSS positioning works. I think `fill` is intended to mean `stretch_to_fit`, because if you set both height and width of an image widget the widget (before your changes) will stretch it. Even with pngs there are lots of cases to think about, svgs possibly even more.

I don't want to hold up this PR with that, so I'll have a look at the rest of it.

src/ui/widgets/Label/label.tsx
```diff
@@ -55,10 +55,17 @@ export const LabelComponent = (
   if (transparent) {
     style["backgroundColor"] = "transparent";
   }
+  if (props.visible) {
```
willrogers, line 58: We're kind of duplicating `useCommonCss()` here, if it's not working there we should change it slightly.

src/ui/widgets/Image/image.tsx
GeorgeSheppard: I have tried that but I'm not sure what the significance of what I'm seeing is. I see what you mean about fill intending to be `stretch_to_fit`, I have changed that on my local version now so can push it, but I thought the way I had done it meant that the previous behaviour would still be applied if the height and width wasn't provided.

src/ui/widgets/Label/label.tsx
GeorgeSheppard: That is a fair point, I will change it to use it.

src/ui/widgets/Image/image.tsx
willrogers: I think you haven't changed the behaviour if width and height aren't provided, but in the tricky 25% case the width **is** provided.

src/ui/widgets/Symbol/symbol.tsx
willrogers: I'm happy to merge this now, because I think it's a step forward anyway, but I agree that this looks fishy. Can you put an example on the widget test page that demonstrates the problem you're solving?

[Another example where just being able to look at your PC would probably save some time.]

src/ui/widgets/Image/image.tsx
GeorgeSheppard: ohhhh I understand, I am a little confused as to why width properties were added to files such as testWidgets.json when there was no width property in the image component though?

willrogers: Positioning is handed in `Widget` as opposed to the individual widgets to avoid having to repeat the code everywhere. To reiterate though, I'm not sure that we've got the best way of handling these positions in all cases so I don't think we should get stuck on this right now. It can be fixed later.

src/ui/widgets/Symbol/symbol.tsx
GeorgeSheppard: I can place a screenshot here so it remains somewhere accessible.
Effectively the first part of the image is with stretchToFit false, the image is the right size and the text is just below it.
The second part of the image is with stretchToFit true, which causes the image to expand (like it does in CS-Studio), but means that the text is pushed further down because it is placed after the image.
Third is the desired behaviour from CS-Studio, where the image is stretched but the text stays where it was before, this is why this line is added, as I cannot think of a proper way of doing it.
StretchToFit

willrogers: OK, thanks, I suggest we merge and move on.

willrogers:
This build is now failing either because of an intermittent test failure or a glitch. Let's merge.

PR #202, opened 2020-10-20
Polyline widget

GeorgeSheppard:


willrogers:
If I'm right, this widget is hard-coded to a cyan horizontal line. In one way I like that you are specifically aiming for the OPI to look the same (i.e. no premature optimisation or generalisation) but I think this is not an especially useful widget as it is.

The CS-Studio polyline widget allows multiple segments. What does the XML fragment that you are converting from look like?

GeorgeSheppard:
I would agree it is not very useful, the reasoning behind that is just that there is only one instance of the polyline object in the synoptic; so there wasn't much to go off in terms of functionality. That being said I can go searching for more instances of it in other beamlines to get a better idea, do you think that is worth doing?

In terms of the XML fragment, it is quite large so I will just say that it starts line 1787 in the synoptic.opi file. There are many other properties, including a large 'scripts' property, but I remember in the first meeting Tom said to ignore those scripts for now. Other than a variable background color there isn't much of interest.

willrogers:
You can always play around with the widget in the OPI editor to get a feel for what options are available. In the end we are looking to maintain compatibility with CS-Studio and opibuilder so we want our widgets to be similar unless we've got a reason otherwise. You're right not to chase down every little feature of every widget though, that isn't good value for our time when we are prototyping.

There is a difficulty with lines because they are unambiguous with absolute positioning but much less clear with relative positioning.

You should use `useCommonCss()` - this will automatically get a bunch of properties including foreground and background colours.

GeorgeSheppard:
That is a good idea I didn't think about doing that. What is opibuilder? Sure will update that.

willrogers:
Right click on the OPI and select "Edit OPI" (and accept the suggestion).  This will take you into the editor view. Here you can drag and drop widgets and change their properties at the right of the screen. It's quite easy to see how the properties map on to the XML representation.

opibuilder is the part of CSS that manages the OPI screens. There are a lot more functions in CSS, databrowser is the one that we use most often at Diamond.

GeorgeSheppard:
That should be all done now, I've added rotations, background color, transparency, and visibility.

willrogers:
I'd call the prop `lineLength` rather than `width` and the widget `Line` rather than `Polyline`, but otherwise we can merge this one.

GeorgeSheppard:
That's an easy one so can change that now.

willrogers:
Good question. Potentially widget width and line length are two different things. What happens if you change it?

GeorgeSheppard:
So if you change it, it ruins the Symbol widget which also has the width property. An alternative is to map the width property in CS-Studio to both `width` and `lineLength` so that you can choose what is more appropriate for the component, but it seems a little wasteful.

willrogers:
Have a look at https://github.com/dls-controls/cs-web-proto/pull/206/files - here I've removed the `width` and `height` props from the symbol widget and I think it's working OK.

willrogers:
Again, no big deal, let's merge this one.

GeorgeSheppard:
My only thinking is that if we change the name of a property a large amount from CS-Studio to this that it will be harder to figure out later when things go wrong. e.g. if later there is another component that requires the CS-Studio width prop, and we have mapped it to lineLength, we just have the same problem as now but for that component instead. Basically I am a little against changing the mappings until I have the list of all properties used in CS-Studio and what components use them.

willrogers:
I don't think that's what's happening. If you look at `opiParsePosition`, that's where `width` and `height` are being parsed. We don't generally use the width or height explicitly in a widget, we use the size of container to size the widget.

Now, in the case of the polyline widget your conversion is only in luck because the polylines you are trying to convert happen each to be a horizontal line from the left to the right of the widget. That's why line length corresponds to widget width. If you happened to be trying to convert a vertical line I thnk you would still produce a horizontal line from the code as it is (unless the line was a horizontal line that had been rotated, which it most likely wouldn't be). The actual information about what line is drawn is in the `points` prop and that is being ignored.

Anyway, these are tricky details for another time.

GeorgeSheppard:
Ok that is my lack of understanding of the parser letting me down there, your point makes sense.

PR #203, opened 2020-10-23
Checkbox

GeorgeSheppard:


src/ui/widgets/Checkbox/checkbox.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  describe("user input changes properties", (): void => {
+    it("clicking on checkbox toggles it", (): void => {
+      const wrapper = mount(
+        <CheckboxComponent {...({ label: "stuff" } as any)} readonly={true} />
+      );
+
+      const findComponent = (): any => wrapper.find(MaterialCheckbox);
+      expect(findComponent().props().checked).toBe(false);
+
+      // Method 1
```
GeorgeSheppard, line 106: @willrogers, would appreciate some help in getting this final test to work, each of the 3 methods I have come up with aren't working and I'm not sure why.

willrogers:
Here I suggest you look at the tests that use 'react testing library'. I started using this more recently and I think it is a better way of creating the tests.

The readback (simple) and tabcontainer (more complex - see `changes tabs on click`) widget tests use this.

GeorgeSheppard:
I've changed tests wherever possible to use react-testing-library, I didn't find it was possible/easily possible to test props using it so I left those tests using react-test-renderer.

willrogers:
Interesting that you've used a Material checkbox here, I think a regular checkbox would have been simpler.

willrogers:
Also, I suggest you keep adding new widgets to the widget test page so we can see the examples.

src/ui/widgets/Checkbox/checkbox.tsx
```diff
@@ -0,0 +1,96 @@
...
+        style={{
+          width: `${width}px`,
+          height: `${height}px`
+        }}
+      />
+    </div>
+  );
+};
+
+const LedWidgetProps = {
```
blondejamtart, line 87: This should probably be renamed to CheckboxWidgetProps (unless I'm missing something)

blondejamtart (changes requested):

src/ui/widgets/Checkbox/checkbox.tsx
GeorgeSheppard: Completely missed that you are right.

GeorgeSheppard:
If you could hold off on merging this for a bit, I think Will is right (I was expecting there to be more benefits of using material ui but it has just added complexity to it). I will try and rework it when I get time this week.

GeorgeSheppard:
So reworking the checkbox to use material-ui made me remember why I switched to it in the first place, it is unreasonably complicated to simply align the text with the checkbox, and as material-ui was already being used elsewhere I thought it would be the simplest way of getting it done without wasting my time.

willrogers:
Closed in favour of #210 .

PR #205, opened 2020-10-26
Colours

willrogers:
I found that the `Color` object became quite complex and it didn't really improve things over this simpler implementation.

@GeorgeSheppard

### DIFF
--- a/src/ui/widgets/Shape/shape.tsx
+++ b/src/ui/widgets/Shape/shape.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { Widget } from "../widget";
+import { useCommonCss, Widget } from "../widget";
 import { WidgetPropType } from "../widgetProps";
 import { registerWidget } from "../register";
 import {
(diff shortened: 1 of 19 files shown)


src/ui/widgets/Shape/shape.tsx
```diff
@@ -23,18 +23,14 @@ const ShapeProps = {
...
 ): JSX.Element => {
-  const newStyle: any = {
+  const style = {
+    ...useCommonCss(props),
     width: props.shapeWidth ? props.shapeWidth : "100%",
     height: props.shapeHeight ? props.shapeHeight : "100%",
     borderRadius: props.shapeRadius ? props.shapeRadius : "",
-    transform: props.shapeTransform ? props.shapeTransform : "",
-    ...props.border?.css()
+    transform: props.shapeTransform ? props.shapeTransform : ""
```
GeorgeSheppard, line 31: Very minor but I think it's nice to use the [coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) operator, saves having to write props.shapeTransform twice.
`props.shapeTransform ? props.shapeTransform : ""` -> `props.shapeTransform ?? ""`

GeorgeSheppard (approved):

PR #206, opened 2020-10-29
Image+symbol

willrogers:
@GeorgeSheppard these are some ideas about the image and symbol widgets - see what you think

I think the image now behaves more like the CS-Studio one, and you can use the labelPosition property on the symbol in the same way as the CS-Studio one as well.

The failing test is interesting because it highlights what I said about react-testing-libary. The test checks for a specific prop, but the failure in the test (the prop is `undefined` not `true`) doesn't actually affect what's shown on the screen, it's really a false test failure.

React testing library encourages you to check what's actually rendered - in this case whether the text is actually rendered - so wouldn't fail in this case where the internals have changed but the behaviour hasn't.


GeorgeSheppard (approved):
Happy with that, especially not hard coding the stretchToFit functionality.

GeorgeSheppard:
Will have a look into what you mean about testing, not sure without looking at the tests again.

willrogers:
Here's a blog post:

https://kentcdodds.com/blog/testing-implementation-details

willrogers:
@GeorgeSheppard should be able to fix the test on this one.

PR #207, opened 2020-10-30
Plot widget

blondejamtart:


GeorgeSheppard (commented):
I like the look of it, and it's very responsive. Think there are build errors currently, and it would be nice to have some tests, happy to do those though as I am trying to practice my testing so let me know if you want me to do them.

blondejamtart:
If you're happy to do them probably better since I have yet to write any tests for this project so I'm not familiar with the frameworks

willrogers:
This is failing on node v10. https://github.com/plotly/react-plotly.js/issues/135#issuecomment-501399736 suggests that people do see this and that node had its default heap size increased in v12.

Comments on that ticket also suggest using `plotly.js-basic-dist` might help.

aawdls:
A new plot widget prototype was implemented in https://github.com/dls-controls/cs-web-lib/pull/15

PR #208, opened 2020-10-30
Coniql.ts build failure

GeorgeSheppard:
Commonly Travis is failing to build a PR due to an error in coniql.ts where the code attempts to access the `.unsubscribe()` method on an undefined object (this error can be seen on the image+symbol open PR). I believe @willrogers attempted a fix but there still seems to be problems. This isn't meant to be permanent just a way to fix the build until we have a better solution.

### DIFF
--- a/src/connection/coniql.ts
+++ b/src/connection/coniql.ts
@@ -262,8 +262,10 @@ export class ConiqlPlugin implements Connection {
       log.error("Websockect client disconnected.");
       for (const pvName of Object.keys(this.subscriptions)) {
         if (this.subscriptions.hasOwnProperty(pvName)) {
-          this.subscriptions[pvName].unsubscribe();
-          delete this.subscriptions[pvName];
+          if (this.subscriptions[pvName]) {
+            this.subscriptions[pvName].unsubscribe();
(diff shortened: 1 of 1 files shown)


GeorgeSheppard:
My guess is that iterating over an object while deleting properties off of that object has some unexpected behaviour. If you have a better fix feel free to change my code.

src/connection/coniql.ts
```diff
@@ -261,8 +261,15 @@ export class ConiqlPlugin implements Connection {
     this.wsClient.onDisconnected((): void => {
       log.error("Websockect client disconnected.");
       for (const pvName of Object.keys(this.subscriptions)) {
-        this.subscriptions[pvName].unsubscribe();
```
blondejamtart, line 264: maybe we could just replace this.subscriptions[pvName].unsubscribe() with this.subscriptions[pvName]?.unsubscribe();
Your fix seems fine though and probably nice to provide some logging like you have

GeorgeSheppard: I agree that would be nicer, but I think if we aren't going to solve the underlying problem we should log it.

PR #209, opened 2020-11-04
Linking container

GeorgeSheppard:
A linking container effectively is an embedded display, but with a groupbox wrapping the display when specified. I have added new .opi files for the linking container on the synoptic, and made minor modifications to other components.

@blondejamtart I have a few things for you to look at if you don't mind. Before @willrogers left after I said what a linking container was he said 'That sounds to me like it should be an embedded display, I don't think we need another widget.'. I had problems just using the embedded display because the embedded display requires an object passed in that isn't created in opiParser, also some of the linkingContainers require borders with labels on (a groupBox) and some don't. It could be I just didn't explain what a linkingContainer was properly and therefore Will was wrong, but I also could be missing something and I could just use an embedded display.
The other thing is I am not sure what an effective way of testing it is, as it is quite hard to test in isolation because the embeddedDisplay is a subcomponent to the linkingContainer. I have tried mocking the embeddedDisplay but I couldn't get it to work. I should add that I could use shallow rendering, but Will has said before that that normally isn't a very good way of doing it.


src/ui/widgets/GroupBox/__snapshots__/groupBox.test.tsx.snap
```diff
@@ -1,13 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`<GroupingContainerComponent /> it matches the snapshot 1`] = `
-<fieldset>
+<fieldset
+  style={
+    Object {
+      "backgroundColor": undefined,
```
blondejamtart, line 7: I guess the background colour prop isn't defined in the snapshot test?
Probably a good idea to do so

GeorgeSheppard: Good idea, done.

PR #210, opened 2020-11-09
Simpler checkbox

willrogers:
@GeorgeSheppard @blondejamtart how about this?


GeorgeSheppard:
I like the look, my one criticism would be that it only toggles if you click the text not the actual textbox. Looking through a lot of the .opi's I have been finding widgets without properties that should really be defined (labels without the text to put in them for example), so in case there is a checkbox without text it would be good if it toggled from the checkbox. Not sure why it doesn't work when clicking the checkbox because the onClick function is wrapping the whole form element.

willrogers:
That's weird - it works for me clicking either the text or the checkbox itself.

GeorgeSheppard:
Hmm, that might be a firefox issue then.

willrogers:
It's working for me on Firefox, Chrome and Edge on Windows. I guess you are Firefox on Linux.

GeorgeSheppard:
@willrogers, seems to be an entirely linux problem with the checkbox not toggling when clicked on directly (instead of the text) (I checked chrome and firefox and had the same issue on both). I think it's worth merging in as I have only seen one checkbox so it's most likely not worth the time. If we care more about clicking on it directly then my more complicated checkbox could be merged instead, but personally I think the simplicity is worth this small drawback (at least for now).

I made a possibly controversial decision to delete the test associated with the checkbox because it is a pretty simple component and the test relied on specific naming of the props. I'd like to think my new javascript hero Kent Dodds would agree with that decision but your opinion is welcome.

willrogers:
I agree regarding the test, I'm puzzled about the behaviour but let's merge ad move on.

blondejamtart:
I've tried it briefly in chromium v86.0.4240.75 on linux (ubuntu) and it seems to work perfectly; I suspect it's an issue with out-of-date versions on the Diamond RHEL7 system

willrogers:
The checkbox is working fine. On the BL21I synoptic there is a transparent label over the top of the checkbox part that is intercepting the clicks.

Try the one on the widget test page instead.

GeorgeSheppard:
That is a good spot, just checked it works well.

PR #213, opened 2020-11-17
WIP: Opi view

GeorgeSheppard:
Expanding the current synoptic view to include two new displays 'main' and 'details'. Clicking on Symbol widgets open a file (if one is tied to that widget) in their respective display.


src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -150,6 +155,19 @@ export function opiParseActions(
     actions: []
   };
```
GeorgeSheppard, line 157: Is this a reasonable way to handle where the new page should open?

src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -176,6 +194,22 @@ export function opiParseActions(
...
         });
+      } else if (type === OPEN_DISPLAY) {
+        processedActions.actions.push({
+          type: OPEN_DISPLAY,
+          dynamicInfo: {
+            name: action.path._text,
+            location: descriptionToLocation(action),
+            description:
+              (action.description && action.description._text) || undefined,
+            file: {
```
GeorgeSheppard, line 205: As the TODO says, I feel like the new pages aren't connecting to coniql because they don't receive their macros but am unsure how to correct this.

src/ui/widgets/widgetActions.ts
```diff
@@ -25,7 +26,10 @@ export interface DynamicAction {
     | typeof OPEN_PAGE
     | typeof CLOSE_PAGE
     | typeof OPEN_TAB
-    | typeof CLOSE_TAB;
+    | typeof CLOSE_TAB
+    // TODO: : OPEN_DISPLAY is the keyword used in .opi files, but operates like
+    // OPEN_TAB in cs-web-proto, probably a more efficient way to combine them both
+    | typeof OPEN_DISPLAY;
```
GeorgeSheppard, line 32: Here I'm unsure if one of the existing DynamicAction properties is what I need to use, currently I copied the action for `OPEN_TAB` as I didn't want to tie the `OPEN_DISPLAY` (the property I made) to something like open tab that may be changed in the future. Thoughts would be good.

GeorgeSheppard:
Would be easiest to not review until #212 is merged so that the files changed on this branch aren't so clogged up. Would appreciate some feedback on the current state of this branch, it's not done yet. The most important thing that needs to be done is getting the new displays to connect to coniql, which I am not sure how to fix at the moment so would appreciate some thoughts on that.

src/ui/widgets/widgetActions.ts
```diff
@@ -98,6 +102,12 @@ export const getActionDescription = (action: WidgetAction): string => {
       } else {
         return `Close tab ${action.dynamicInfo.name}`;
       }
+    case OPEN_DISPLAY:
+      if (action.dynamicInfo.description) {
+        return action.dynamicInfo.description;
+      } else {
+        return `Open display ${action.dynamicInfo.name}`;
+      }
```
GeorgeSheppard, line 110: Currently have made a copy of `OPEN_TAB` (which I'm sure whether it serves the same purpose or not, and called it `OPEN_DISPLAY` instead. Is there an existing DynamicContent property that has the functionality these new pages want? I couldn't find any documentation on the existing properties so wasn't sure.

src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -150,6 +155,19 @@ export function opiParseActions(
...
+    switch (description) {
+      case "Open component screen":
+        return "main";
+      case "Open detail screen":
+        return "details";
+      default:
+        return "details";
+    }
+  };
+
```
GeorgeSheppard, line 170: Is this a good way of mapping the new pages to either the `main` display or the `details` display. I feel like it's a bad way of doing it but can't think of another way.

src/ui/widgets/EmbeddedDisplay/opiParser.ts
```diff
@@ -176,6 +194,22 @@ export function opiParseActions(
...
+            description:
+              (action.description && action.description._text) || undefined,
+            file: {
+              path: action.path._text,
+              // TODO: Should probably be accessing properties of the element here
+              macros: {},
+              defaultProtocol: "pva"
+            }
+          }
+        });
```
GeorgeSheppard, line 212: The reason the new pages aren't connecting to coniql, could be because I am only assigning a blank object to the macros, but am not sure how to access the correct macros from this point in the code.

src/ui/widgets/EmbeddedDisplay/opiParser.ts
willrogers: There must be a way that CS-Studio determines where the screens are opening. I think it might be `<mode>` but I'd have to check.

willrogers:
I've been looking at this this morning, and I've got a branch with a couple of big changes that resolve most of this. To summarise:

* instead of having a special linking container widget, use the embedded display and add the groupbox container into what's rendered if the border style is groupbox. Ideally we would allow any widget to have the groupbox border, but it's a little complicated. The embedded display widget handles all the macros already, so that fixes the connections.
* have a 'complex parser' to parse the opi file details to pass to the embedded display

I've then tweaked things here and there to make it look a bit better.

src/ui/widgets/widgetActions.ts
willrogers: I'll have a look. Do we not want to use `OPEN_TAB`?

src/ui/widgets/EmbeddedDisplay/opiParser.ts
willrogers: Yes, it's messy. You might find `replace` or `mode` props. I think what we can do for now is: if `mode` is not present or is 1, open in main; if `mode` is 3 open in details.

src/ui/widgets/widgetActions.ts
GeorgeSheppard: I'm not sure what we want to use, I didn't know whether one of these actions was designed specifically with this purpose in mind or whether a new one is needed.

willrogers: No, I think we want `OPEN_TAB` here.

src/ui/widgets/EmbeddedDisplay/opiParser.ts
GeorgeSheppard: Ah good spot, I think that would make sense just looking through now, 3 seems to be for the details pane.

GeorgeSheppard:
Those changes sound very useful and I agree embedding the groupbox into the embedded display would be good. Complex parser solves the problem I was having that made me want to create the linking container (the embedded display requires a file object that wasn't created in the parser), so that sounds good too.

I will keep this branch in the background for now then until the changes you talk about are merged in, then can finish it off.

willrogers:
OK, I'll tidy up what I have and push it. We will probably want to squash the changes into one commit in the end.

willrogers:
I've just pushed that, have a look.

src/ui/widgets/EmbeddedDisplay/embeddedDisplay.tsx
```diff
@@ -106,21 +118,12 @@ export const EmbeddedDisplay = (
...
-        ? "scroll"
-        : "visible";
-
     component = widgetDescriptionToComponent({
       type: "display",
       position: props.position,
       border:
         props.border ?? new Border(BorderStyle.Line, new Color("lightgrey"), 1),
-      overflow: overflow,
+      overflow: props.scroll ?? props.scroll ? "scroll" : "visible",
```
GeorgeSheppard, line 126: I may be missing something but can this not be simplified to:
`props.scroll ? "scroll": "visible"`, I don't think we need the ??.

src/ui/widgets/Symbol/symbol.tsx
```diff
@@ -102,6 +102,9 @@ export const SymbolComponent = (props: SymbolComponentProps): JSX.Element => {
     }
   }
 
+  const cursor =
+    props.actions && props.actions.actions.length > 0 ? "pointer" : "auto";
+
```
GeorgeSheppard, line 107: nice touch, didn't know you could do that, I think could be simplified to
`props.actions?.actions?.length > 0 ? "pointer": "auto"` though

GeorgeSheppard:
I can make both the changes I have put above while I add your suggestions from earlier so don't worry about doing them. I'll squash everything at the same time.

src/ui/widgets/Symbol/symbol.tsx
GeorgeSheppard: Ah I assume the reason you didn't do this was because of the typescript error of `this could be undefined`.

src/ui/widgets/widgetActions.ts
GeorgeSheppard: In this case can we rename `OPEN_TAB` to `OPEN_DISPLAY` as "OPEN_DISPLAY" is how it comes in on the props from the .opi. Also line 191 of opiParser would be more consistent with the rest of the if else statements.

GeorgeSheppard: Just looked at the background code, and it would be a pain to change all instances of tab to display, so perhaps could just be changed in widgetActions, as widgetActions makes the calls to the fileContext.

willrogers: I don't particularly feel the need to change this one as I think it makes sense for our app. `OPEN_PAGE` will open the file in any `DynamicPage` with the right location and `OPEN_TAB` will open the file in any `DynamicTabs` with the right location.

GeorgeSheppard: That's fine, will just have to map the `OPEN_DISPLAY` to `OPEN_TAB` then.

GeorgeSheppard:
@willrogers everything should be sorted now. Have checked with NoMachine and all the new embedded displays are connecting correctly.

PR #216, opened 2020-12-01
Devices updated

GeorgeSheppard:
@willrogers I thought it might be worthwhile opening a pull request now to get the infrastructure for devices all merged in. The device widget is still basic, but it collects the device json and sends it down to the component.


willrogers:
Thanks, I'm having a look this morning.

willrogers:
It is a complicated part of the code, isn't it?

I see one pretty big simplification. If you look at https://github.com/dls-controls/coniql/blob/master/coniql/schema.gql, you see that you can get a channel either with a query or a subscription, but you can only get a device with a query.  I think Tom intends that devices don't update.

So if a widget needs a device, we just execute the query and store the result. We don't need to unsubscribe or handle updates. I think the only addition to `Connection` would be a new method `getDevice()`. The device would contain channels and we would need to subscribe to those.

What do you think? @thomascobb any thoughts?

src/connection/coniql.ts
```diff
@@ -234,16 +238,45 @@ const PV_MUTATION = gql`
...
+            layout
+            children {
+              name
+            }
+          }
+        }
+      }
+    }
+  }
+`;
```
thomascobb, line 266: This should be all you ever have to do, on connection to coniql. And it is static, so can't change without restarting coniql, so you never need to refetch it. This means no subscribing or unsubscribing, the first time you are asked about the contents of a Device you run this query, then subscribe to the Channels give.

thomascobb: So yes, what @willrogers said is correct

GeorgeSheppard:
Ah, well that is quite a big misunderstanding on my part then. Yes it is quite complicated, but I have wrapped in a few different things into the pull request which may exaggerate the difficulty.

GeorgeSheppard:
It's not all wasted time as I understand it now so I will change that tomorrow then. @thomascobb the current M1Yaw device example doesn't return channel names but returns string current values, is that what it is meant to be doing if instead of subscribing to the device we should be subscribing to the channels given? The current response looks something like this just for context to Will and others
```
{
          "label": "Demand",
          "child": {
            "value": {
              "string": "0.005100"
            }
          }
        },
        {
          "label": "Readback",
          "child": {
            "value": {
              "string": "0.005100"
            }
          }
        },
```

GeorgeSheppard:
Closed in favour of #217.

PR #217, opened 2020-12-04
Device query

GeorgeSheppard:
@willrogers, some initial feedback on what you think of the backend work would be helpful. Will add new tests once I know this is the right direction.

### DIFF
--- a/src/connection/coniql.ts
+++ b/src/connection/coniql.ts
@@ -22,7 +22,9 @@ import {
   ValueChangedCallback,
   nullConnCallback,
   nullValueCallback,
-  SubscriptionType
+  SubscriptionType,
+  DeviceQueriedCallback,
+  nullDeviceCallback
 } from "./plugin";
 import { SubscriptionClient } from "subscriptions-transport-ws";
 import {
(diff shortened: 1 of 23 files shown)


src/connection/coniql.ts
```diff
@@ -317,10 +348,12 @@ export class ConiqlPlugin implements Connection {
 
   public connect(
     connectionCallback: ConnectionChangedCallback,
-    valueCallback: ValueChangedCallback
+    valueCallback: ValueChangedCallback,
+    deviceQueried: DeviceQueriedCallback
```
GeorgeSheppard, line 352: I dislike having to add an extra function here to handle the device query but wasn't sure there was a better way of doing it. Using the second function to handle device values as well seemed a bit opaque to me.

willrogers:
Yes, this looks more like what I expected.

src/connection/coniql.ts
willrogers: This looks ok to me.

GeorgeSheppard:
@willrogers this should be the backend for devices done now hopefully

src/ui/hooks/utils.ts
```diff
@@ -16,6 +17,16 @@ export function pvStateSelector(
   return results;
 }
 
+export function deviceSelector(device: string, state: CsState): DType {
+  return state.deviceCache[device];
+}
+
+export function deviceComparator(before: DType, after: DType): boolean {
```
willrogers, line 24: I'm not so sure about this one. If I am correct, `deviceComparator()` will be called each time the store updates, then anything using `useDevice()` will be re-rendered each time the store updates.

I'm not certain, so I propose we leave it as it is and then we can check our understanding.

willrogers:
This looks fine to me. Would you like to merge, or continue working on this branch?

src/ui/hooks/utils.ts
GeorgeSheppard: Ahh interesting, my understanding was that redux passes in the previous value for a specific device (found using the deviceSelector first), and then just compares the old with the new. I'll have a look into it.

GeorgeSheppard:
Merging would be good, I'm conscious of the refactor you want to merge in so don't want this branch to get too outdated.

PR #221, opened 2020-12-07
Widget refactor

willrogers:
@GeorgeSheppard @blondejamtart a substantial refactor with probably no change to what is rendered. The aim was to get rid of the 'widget wrappers' and the additional divs that they added.

Widget and its child ConnectingComponent are a bit more complicated, but other complexity (e.g. recursiveWrapping) has been removed.

See what you think.


GeorgeSheppard (approved):
Looks good to me, really like the right click functionality.

PR #224, opened 2020-12-16
Device parsing

GeorgeSheppard:
- Parsing to turn coniql device response into components
- throttleMiddleware bug was causing the queue not to flush, fixed this
- Tweaks to some of the infrastructure (hooks, comparator/selector)
- Modified motor_embed.opi to load new device.opi instead of old motor_detail.opi, device.opi uses parent macros to define the device to query from coniql
- detailPanel gets mapped to device

### DIFF
--- a/src/redux/throttleMiddleware.test.ts
+++ b/src/redux/throttleMiddleware.test.ts
@@ -53,7 +53,6 @@ describe("throttleMidddlware", (): void => {
   });
   it("dispatches ValuesChanged when receiving ValueChanged", (): void => {
     const middleware = throttleMiddleware(updater);
-    updater.ready = true;
     // nextHandler takes next() and returns the actual middleware function
     const nextHandler = middleware(mockStore);
     const mockNext = jest.fn();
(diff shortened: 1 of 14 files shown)


willrogers:
Thanks @GeorgeSheppard - last day on the project so let's get this merged.

Why did you change from `motor_embed.opi` to `device.opi`? What's the difference?

GeorgeSheppard:
@willrogers, so the `motor_embed.opi` had extra widgets on it, `device.opi` is just a stripped down `.opi` with only a device widget on it. I figured as the plan was to just display the device widget in the detail pane that it was worth creating a simplified `.opi` for that purpose.

willrogers:
If you open the device page when you're not connected to coniql, it crashes. I guess we need some exception handling.

src/redux/throttleMiddleware.test.ts
```diff
@@ -53,7 +53,6 @@ describe("throttleMidddlware", (): void => {
   });
   it("dispatches ValuesChanged when receiving ValueChanged", (): void => {
```
willrogers, line 54: I'm not really sure what this test is trying to show now.

willrogers:
Mainly looks good, I've made a couple of comments but let's try to merge it this afternoon.

src/redux/throttleMiddleware.test.ts
GeorgeSheppard: Have changed it now so the test manually triggers the queue flushing.

willrogers:
One more comment: in `deviceParser` you construct a string, then later use `parseJson` to turn it back into objects. Would it not be more efficient to create the objects then use `parseObject` instead?

GeorgeSheppard:
@willrogers I'm glad you pointed that out I didn't even consider that but it's much cleaner, all changed now.

PR #249, opened 2021-08-12
Allow secure connections to Coniql.

willrogers:


willrogers:
cc @AlexanderWells-diamond

PR #266, opened 2023-05-25
Refactor to use the cs-web-lib library

rjwills28:
I have refactored cs-web-proto to use the library functions from cs-web-lib. There are quite a few changes but most are deleted files and updating imports to cs-web-lib. I have also made a few changes to some of the opi and json files to optimise the display.


abigailalexander:
I ran this locally and it looks good - I did notice however that on the Widget demo page the TabContainer widget is missing. It seems this is something not exported from cs-web-lib. Nothing else appears to be missing though

rjwills28:
@abigailalexander thanks for catching the missing tab container widget. I fixed the export of that in cs-web-lib and have updated the dependency in cs-web-proto to use the latest lib version.
Indeed, some of the opis (including TMBF) do now appear bigger as we have implemented the autoZoomToFit property in cs-web-lib. They now scale up or down to fit the window size. This can be turned off by changing the autoZoomToFit property to false.

abigailalexander (approved):

